PR #38, opened 2021-12-07
fix: conform search provider to warp2 schema

KrisQ:
**Ticket**

[LS-1213](https://nacelle.atlassian.net/browse/LS-1213)

**Changes**

conform search provider + testing to warp2 schema

**Testing**

- `git checkout LS-1213--warp2-search-provider`
- `cd packages/vue`
- `npm run test -- searchProvider.spec.js`


alexcasche (approved):
This looks good!  How do we feel about merging this into `main` vs another `warp2` dedicated branch.  I could see some value in separation here

KrisQ:
> This looks good! How do we feel about merging this into `main` vs another `warp2` dedicated branch. I could see some value in separation here

Good point let's do that instead!

PR #39, opened 2021-12-09
fix: make product provider compatible with warp2

KrisQ:
**Ticket**

https://nacelle.atlassian.net/browse/LS-1212

**Changes**

make product provider, tests and mock compatible with warp2

**Testing**

- `git checkout LS-1213--warp2-product-provider`
- `cd packages/vue`
- `npm run test -- productProvider.spec.js`


alexcasche (approved):
I think this is good until we figure out the best way for fetching data.  Eventually line 58 would need to be removed or replaced with storefront-sdk ```productObject = await nacelleSdk.data.product({ handle });```

PR #40, opened 2021-12-09
chore: adds typecheck for string

alexcasche:
**What This PR Does**
- ensures shopifyCheckout package only passes through `checkoutAttributes` with values that are strings

**Testing**
- `npm run test:ci`

**Notes**
- the `Attribute` and `Metafield` types were left unchanged intentionally so that typescript users would still get type errors if the values passed weren't strings.  The caveat is that there is no test against passing in other values.  Another approach where types are updated can be seen on this branch: 
https://github.com/getnacelle/nacelle-js/tree/LS-1306--null-checkout-attrs

### DIFF
--- a/packages/shopify-checkout/src/utils/metafieldsToCustomAttributes.ts
+++ b/packages/shopify-checkout/src/utils/metafieldsToCustomAttributes.ts
@@ -17,7 +17,9 @@ export default function metafieldsToCustomAttributes({
   }
 
   const customAttributes: Attribute[] = metafields?.reduce((fields, m) => {
-    fields.push({ key: m.key, value: m.value });
+    if (typeof m.value === 'string') {
+      fields.push({ key: m.key, value: m.value });
+    }
 
     return fields;
   }, [] as Attribute[]);
--- a/packages/shopify-checkout/src/utils/metafieldsToCustomAttributes.ts
+++ b/packages/shopify-checkout/src/utils/metafieldsToCustomAttributes.ts
@@ -19,6 +19,10 @@ export default function metafieldsToCustomAttributes({
   const customAttributes: Attribute[] = metafields?.reduce((fields, m) => {
     if (typeof m.value === 'string') {
       fields.push({ key: m.key, value: m.value });
+    } else {
+      console.warn(
+        `Omitting custom attribute "${m.key}" because it has a non-string value.`
+      );
     }
 
     return fields;


packages/shopify-checkout/src/utils/metafieldsToCustomAttributes.ts
```diff
@@ -17,7 +17,9 @@ export default function metafieldsToCustomAttributes({
   }
 
   const customAttributes: Attribute[] = metafields?.reduce((fields, m) => {
-    fields.push({ key: m.key, value: m.value });
+    if (typeof m.value === 'string') {
+      fields.push({ key: m.key, value: m.value });
+    }
```
NWRichmond, line 22: One possibility here would be to stringify non-string values. For example:

```ts
if (typeof m.value === 'string') {
  fields.push({ key: m.key, value: m.value });
} else {
  fields.push({ key: m.key, value: JSON.stringify(m.value) });
}
```

It looks like that would work equally well for numbers, arrays, and objects:

JSON stringify number, array, and object

NWRichmond (commented):
Looks good! I'd like to get your thoughts about whether we should coerce non-strings to strings. Also, would it be worth adding another assertion to validate this behavior, maybe in `packages/shopify-checkout/src/client/client-dom.spec.ts`?

packages/shopify-checkout/src/utils/metafieldsToCustomAttributes.ts
alexcasche: @NWRichmond I considered that.  My thought was merchants might not want to pass along "undefined".  I guess I'm not too opinionated here.

NWRichmond: For sure. Makes me wonder if we should instead:

```ts
if (typeof m.value === 'string') {
  fields.push({ key: m.key, value: m.value });
} else {
  console.warn(`Omitting custom attribute "${key}" because it has a non-string value.`);
}
```

alexcasche: Sounds good to me!  I'll get that into this PR and tested

alexcasche: @NWRichmond since the `metafield.value` is typed as a string, I'm not able to test this easily.  Basically the test fails with Typescript errors.

NWRichmond (approved):
Looks good! Even without testing this new `if () {} else {}` branch, we're still enjoying high code coverage.

LS-1306--null-undefined-metafields-checkout branch code coverage

PR #42, opened 2021-12-16
chore: rm collection provider

alexcasche:
**What This PR Does**
- removes `CollectionProvider` and `useCollectionProvider`.
- starter/reference stores will need updates too but these are blocked until offset pagination is added to API.


KrisQ (approved):
Looks good!

PR #48, opened 2022-01-06
feat(nuxt-starter): rm @nuxt/composition-api

alexcasche:
**What This PR Does**
- swaps `@nuxtjs/composition-api` for `@vue/composition-api`
- updates components to remove any `nuxt-hooks` from `@nuxtjs/composition-api`
- moves logic in `/plugins/globalSetup` to `/layouts/default`

**How To Test**
- `git checkout LS-1417--rm-nuxt-composition`
- `yarn dev`
- check that pages load and state is maintained: `/search`, `/collections/clothing`, `/products/otto-shirt`, `/cart`

**Notes**
- seeing a warning in console on pages that fetch data `Cannot stringify arbitrary non-POJOs NacelleClient` - ex: `/products/otto-shirt`.  Working on sorting this out now


mdarrik:
Looks really good! Little side bonus is that the current version of NuxtBridge requires us to move the globalSetup plugin to the `layout` anyways.

alexcasche:
This route is no longer an option

PR #49, opened 2022-01-07
refactor(starters/nuxt): useAsync

alexcasche:
**What This PR Does**
- swaps `useAsync` for `useFetch`
- removes `onGlobalSetup` plugin, moves logic to `/layouts/default`

**Notes**
- if we are sold on this approach we can merge.  otherwise we can leave open until we have a clear path forward


NWRichmond (commented):
Awesome! I just tested this locally, and it worked great with `dev` and `generate`.

One improvement that I think we could squeeze in alongside these changes is removing `generate.interval` from `nuxt.config.js`. It was needed with an earlier version of `@nuxtjs/composition-api`, but there is apparently no need for it in the most recent version that we're using (it's not mentioned anywhere in [composition-api.nuxtjs.org](https://composition-api.nuxtjs.org/getting-started/gotchas) anymore). When I removed it and `npm run generate && npx serve dist`, pages appeared to be generated as expected, with a huge decrease in build time:

Aurora High Tops PDP built as expected
:

jasonredick (approved):
Looks great and runs great. I'm all for going with this approach. Seems like it's simpler to manage too.

alexcasche:
Closing as we are moving away from `@nuxtjs/composition-api`

PR #51, opened 2022-01-19
feat(starters/nuxt-2): nuxt-2 starter powered by .query

alexcasche:
**Context**
- Ticket:  [LS-1459](https://nacelle.atlassian.net/browse/LS-1459)

**What This PR Does**
- Adds `/starters/nuxt-2` only using Options API
- Fetches data using `@nacelle/storefront-sdk` `query` method

**Notes**
- `/search` has been dropped until we have a method we can get behind (backend search service)

**Testing**
- Clone project
- Add `.env`
```
# Nacelle
NACELLE_STOREFRONT_ENDPOINT="https://storefront.api.development.nacelle.com/graphql/v1/spaces/cky4xgh48965713nxjkv9s75uk"
NACELLE_STOREFRONT_TOKEN="16f0d12a-7990-47dc-a5f0-4bd043e2b6e7"
NACELLE_STOREFRONT_LOCALE="en-US"

# Shopify Checkout (for details, see: https://www.npmjs.com/package/@nacelle/shopify-checkout)
MYSHOPIFY_DOMAIN="pepper-wood-apparel.myshopify.com"
SHOPIFY_STOREFRONT_CHECKOUT_TOKEN="477e3f771eb126e58db487428cd34948"
STOREFRONT_API_VERSION="2022-01"
```
- Check routes `/products/otto-shirt`, `/collections/clothing`, `/cart`
- Attempt a checkout


alexcasche:
@NWRichmond @mdarrik.  When setting this project up I used `create-nuxt-app` to get the latest setup/packages. But now lerna complains about mismatching versions for hoisted packages.  What is the best approach to avoid this when spinning up new starters, etc.

mdarrik:
I've been hitting the same issues with Lerna. It's one of the more frustrating challenges with it. Either we have to remove the `--hoist` flag, or touch a bunch of other dependencies and get them to match. (Or keep new things on old tools).

alexcasche:
> I've been hitting the same issues with Lerna. It's one of the more frustrating challenges with it. Either we have to remove the `--hoist` flag, or touch a bunch of other dependencies and get them to match. (Or keep new things on old tools).

I also assume that depending on what ends up in starters we may intentionally want different versions.  Maybe Next needs a different version than Nuxt, etc

NWRichmond:
> > I've been hitting the same issues with Lerna. It's one of the more frustrating challenges with it. Either we have to remove the `--hoist` flag, or touch a bunch of other dependencies and get them to match. (Or keep new things on old tools).
> 
> I also assume there will be scenarios where we need different versions. Maybe Next starter needs a different version of a package than Nuxt, etc

You can selectively not hoist select packages, e.g. by changing the `boostrap` command to:

```
lerna bootstrap --hoist --strict --nohoist=some-package
```

starters/nuxt-2/components/ProductCard.vue
```diff
@@ -0,0 +1,156 @@
...
+    >
+      <nuxt-img
+        v-if="product.content.featuredMedia"
+        :src="product.content.featuredMedia.src"
+        class="product-card__image"
+      />
+      <div v-else>No Image</div>
+    </nuxt-link>
+    <div class="product-card__main">
+      <h3 v-if="product.content.title" class="product-card__title">
```
mdarrik, line 15: Based on how we're using this in the starter, this should probably be an `h2`. But it's one of those places where components butt up against html since there may be instances where this actually needs to be an `h3`.

starters/nuxt-2/components/ProductCard.vue
```diff
@@ -0,0 +1,156 @@
...
+          ${{ selectedVariant.compareAtPrice }}
+        </div>
+        <div class="product-card__price">${{ selectedVariant.price }}</div>
+      </div>
+      <div
+        v-for="(option, oIndex) in options"
+        :key="oIndex"
+        class="product-card__option"
+      >
+        <label class="product-card__label">{{ option.name }}</label>
```
mdarrik, line 32: label needs a `:for="id"` in order to correctly connect the label to the select

starters/nuxt-2/pages/products/_handle.vue
```diff
@@ -0,0 +1,216 @@
...
+          ${{ selectedVariant.compareAtPrice }}
+        </div>
+        <div class="product__price">${{ selectedVariant.price }}</div>
+      </div>
+      <div
+        v-for="(option, oIndex) in options"
+        :key="oIndex"
+        class="product__option"
+      >
+        <label class="product__label">{{ option.name }}</label>
```
mdarrik, line 25: Same as above:
label needs a :for="id" in order to correctly connect the label to the select

starters/nuxt-2/pages/products/_handle.vue
```diff
@@ -0,0 +1,216 @@
...
+        </select>
+      </div>
+      <!-- eslint-disable vue/no-v-html -->
+      <div
+        v-if="product.content.description"
+        class="product__description"
+        v-html="product.content.description"
+      />
+      <div class="product__quantity">
+        <label class="product__label">Quantity:</label>
```
mdarrik, line 46: Tagging the label for again here

starters/nuxt-2/store/README.md
```diff
@@ -0,0 +1,10 @@
+# STORE
```
mdarrik, line 1: Probably worth replacing some of this with info related to the existing store. Like the lifecycles of actions and things.

mdarrik (commented):
Some small a11y considerations we should take so users start out on the right foot, but otherwise seems okay. 

It also doesn't necessarily have to be part of this PR, but I'd love to see some comments documenting the methods and stuff we have in place so that users taking over have a good sense of what everything is doing. If we think it's out of scope for this, we should definitely make a ticket for it though. Because I think that's a huge missing piece of the starters. 

Did we decide this was going to be a full starter or an example which we can link to? I can't remember if we came to a clear consensus earlier or not.

mdarrik:
> You can selectively not hoist specific packages, e.g. by changing the `boostrap` command to:
> 
> ```
> lerna bootstrap --hoist --strict --nohoist=some-package
> ```

Yeah, I definitely did this for the example with Vue 3. But it sort of becomes cumbersome as we add more and more packages/projects if we can't keep all of their dependencies aligned. And it sort of feels weird to add new packages and simultaneously update a bunch of other packages so that they all have version alignment. 

Lerna's resolution is also kind of brittle. Eg if a package has `^1.0.0` as a dependency, and another has `^1.4.0`, Lerna will complain that they don't match even if handling `^1.4.0` would include `^1.0.0`.

starters/nuxt-2/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Example: `/static/robots.txt` is mapped as `/robots.txt`.
+
+More information about the usage of this directory in [the documentation](https://nuxtjs.org/docs/2.x/directory-structure/static).
+
+### `store`
+
+This directory contains your Vuex store files. Creating a file in this directory automatically activates Vuex.
+
+More information about the usage of this directory in [the documentation](https://nuxtjs.org/docs/2.x/directory-structure/store).
```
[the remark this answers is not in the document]
alexcasche: Sounds like a good idea.  This is just pulled from default nuxt projects.  I think another ticket for this makes sense

starters/nuxt-2/store/README.md
alexcasche: Agreed.  I think a follow-up README.md task for the entire repo makes sense

alexcasche:
@mdarrik all updates should be good now, minus the README.  I think that should be a separate ticket

jasonredick (approved):
This looks great to me! I'll echo Darrik's comments about the improved README and adding comments to the methods we are using, that's a great idea.

starters/nuxt-2/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@nacelle/storefront-sdk": "^1.0.2",
+    "core-js": "^3.19.3",
+    "fuse.js": "^6.5.3",
+    "idb-keyval": "^6.1.0",
+    "nuxt": "^2.15.8",
+    "uuid": "^8.3.2",
+    "vue": "^2.6.14",
+    "vue-server-renderer": "^2.6.14",
+    "vue-template-compiler": "^2.6.14",
+    "webpack": "^4.46.0"
```
NWRichmond, line 25:
```suggestion
```

`vue-server-renderer`, `vue-template-compiler`, and `webpack` look like extraneous dependencies. For reference, here's the `package.json` of a fresh Nuxt 2 project created with `npm init nuxt-app`, and some pretty minimal CLI options chosen:

```json
{
  "name": "new-nuxt-app",
  "version": "1.0.0",
  "private": true,
  "scripts": {
    "dev": "nuxt",
    "build": "nuxt build",
    "start": "nuxt start",
    "generate": "nuxt generate",
    "lint:js": "eslint --ext \".js,.vue\" --ignore-path .gitignore .",
    "lint": "npm run lint:js"
  },
  "dependencies": {
    "@nuxtjs/pwa": "^3.3.5",
    "core-js": "^3.15.1",
    "nuxt": "^2.15.7"
  },
  "devDependencies": {
    "@babel/eslint-parser": "^7.14.7",
    "@nuxtjs/eslint-config": "^6.0.1",
    "@nuxtjs/eslint-module": "^3.0.2",
    "eslint": "^7.29.0",
    "eslint-config-prettier": "^8.3.0",
    "eslint-plugin-nuxt": "^2.0.0",
    "eslint-plugin-vue": "^7.12.1",
    "prettier": "^2.3.2"
  }
}
```

package json npm init nuxt-app

starters/nuxt-2/pages/index.vue
```diff
@@ -0,0 +1,45 @@
...
+      <a href="https://docs.getnacelle.com/nuxt/intro-nuxt.html" target="none"
+        >visit the Nacelle documentation.</a
+      >
+    </p>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'IndexPage'
```
NWRichmond, line 19: This might be clearer:

```suggestion
  name: 'HomePage'
```

It would also be consistent with classes used in the template.

starters/nuxt-2/pages/products/_handle.vue
```diff
@@ -0,0 +1,221 @@
...
+      </h1>
+      <div class="product__prices">
+        <div v-if="selectedVariant.compareAtPrice" class="product__compare">
+          ${{ selectedVariant.compareAtPrice }}
+        </div>
+        <div class="product__price">${{ selectedVariant.price }}</div>
+      </div>
+      <div
+        v-for="(option, oIndex) in options"
+        :key="oIndex"
```
NWRichmond, line 22: Might be safer to use `option.name` here.

starters/nuxt-2/store/checkout.js
```diff
@@ -0,0 +1,34 @@
...
+      if (checkoutId) {
+        const checkout = checkoutClient.get({
+          id: checkoutId
+        });
+        if (checkout?.completed) {
+          await del('checkoutId');
+          commit('cart/clearCart', false, { root: true });
+        }
+      }
+    } catch {}
```
NWRichmond, line 18: Does this really need to be in a `try/catch`? If so, let's do something in the `catch`.

starters/nuxt-2/store/checkout.js
```diff
@@ -0,0 +1,34 @@
...
+      const cartItems = rootState.cart.lineItems.map((lineItem) => ({
+        quantity: lineItem.quantity,
+        variantId: lineItem.variant.id
+      }));
+      const checkoutData = await checkoutClient.process({ cartItems });
+      await set('checkoutId', checkoutData.id);
+      if (checkoutData.url) {
+        window.location.href = checkoutData.url;
+      }
+    } catch {}
```
NWRichmond, line 32: We should either:
- Do something in the `catch` block, or:
- Remove the `try/catch`, and add a `.catch` to `.process`, and do something in the `.catch`.

When `catch`ing an error in `checkoutClient.process`, we could at least `console.error(err)`. The merchant developer may want to trigger a modal,  but I think it would be too heavy-handed/opinionated to implement that here (please feel free to disagree!).

starters/nuxt-2/utils/buildRoutes.js
```diff
@@ -0,0 +1,33 @@
...
+  `;
+
+  const response = await client.query({ query: QUERY });
+  return [
+    ...response?.products?.map(
+      (product) => `/products/${product.content.handle}`
+    ),
+    ...response?.productCollections?.map(
+      (productCollection) => `/collections/${productCollection.content.handle}`
+    )
```
NWRichmond, line 31: Minor nitpick, but after testing with the Nacelle Storefront API, it looks like the result of these queries should always be `[]`, and not falsey (`null`), so we could lose the `?` before the `.map`s.

productCollections nonexistent

NWRichmond (changes requested):
Nicely done!

I am really enjoying the use of `.query` here; I suspect that it will be a long-term win for merchant developers.

I made some suggestions, but they are focused on just adding some icing to the cake. Overall, `starters/nuxt-2` is in very good shape.

### Lerna

**UPDATE:** the issue described below is resolved as of https://github.com/getnacelle/nacelle-js/pull/51/commits/da0e24685a27818da7212c297dc4b20fffc355eb.

When running `npm run bootstrap` in the monorepo root, I'm seeing:

lerna dependency issues

The underlying issue is that there are multiple versions of the same package, and Lerna (when running `bootstrap` with the `--strict` flag) is telling us that that's an issue. But many of these dependencies are straightforward to resolve. For instance:

nacelle-js eslint versions

In this case, we can install `eslint@8.4.1` in the monorepo root.

I've created a branch, [`LS-1459-dependencies`](https://github.com/getnacelle/nacelle-js/compare/LS-1459--nuxt-2-starter-query-method...LS-1459-dependencies), which resolves the dependencies. We coulg merge that branch into this branch.

### [`jsconfig.json`](https://code.visualstudio.com/docs/languages/jsconfig)

To add one more suggestion, it might be wise to add a `jsconfig.json` to this project. In [`create-nuxt-app`](https://github.com/nuxt/create-nuxt-app), it's an option that's recommended to support the VSCode x JavaScript development experience:

npm init nuxt-app options jsconfig

When I selected that option, `create-nuxt-app` produced the following `jsonconfig.json`:

```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "~/*": ["./*"],
      "@/*": ["./*"],
      "~~/*": ["./*"],
      "@@/*": ["./*"]
    }
  },
  "exclude": ["node_modules", ".nuxt", "dist"]
}

```

alexcasche:
@NWRichmond change requests should be good to go.  I'm going to hold off on the README.md.  They will likely include links out to the docs, so we should get those sorted out first.

NWRichmond:
> @NWRichmond change requests should be good to go. I'm going to hold off on the README.md. They will likely include links out to the docs, so we should get those sorted out first.

Thanks @alexcasche! The only outstanding issue I'm seeing is https://github.com/getnacelle/nacelle-js/pull/51/files#r790352333.

starters/nuxt-2/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+- [`@nacelle/storefront-sdk`](https://www.npmjs.com/package/@nacelle/storefront-sdk) - Nacelle Storefront SDK, providing methods to query data from your Nacelle space.
+- [`@nacelle/shopify-checkout`](https://github.com/getnacelle/nacelle-js/tree/main/packages/shopify-checkout) - Helpful utilities for handling Shopify checkouts.
+
+### Nuxt.js
+
+To learn more about Nuxt.js, take a look at the following resources:
+
+- [Nuxt.js Documentation](https://nuxtjs.org/docs) - learn about Nuxt.js features and API.
+- [Learn Nuxt.js](https://nuxtjs.org/tutorials) - a setup of helpful Nuxt.js tutorials.
```
NWRichmond, line 69:
```suggestion
- [Learn Nuxt.js](https://nuxtjs.org/tutorials) - a set of helpful Nuxt.js tutorials.
```

NWRichmond (approved):
Excellent!

Everything works as expected on the dev server & build.

https://user-images.githubusercontent.com/5732000/151385555-8400c2c5-e3f9-4e26-8f9f-9767c37ad359.mov

Thanks so much for this PR!

PR #52, opened 2022-01-24
chore: use common dependencies

NWRichmond:
This PR resolves dependency disagreements that were causing problems with `lerna bootstrap --hoist --strict` in the `LS-1459--nuxt-2-starter-query-method` branch (#51).

lerna bootstrap LS-1459-dependencies


alexcasche (approved):
LGTM thanks Nick!

PR #62, opened 2022-01-27
refactor: nuxt project directories

alexcasche:
**What This PR Does**
- moves `starters/nuxt` starter to `examples/nuxt-2-composition-api`
- renames `starters/nuxt-2` to `starters/nuxt`

**Why This PR Is Needed**
- we should only have 1 starter per meta-framework.  It should always be whatever is `stable`.  All other starter-like projects should be in examples directory


examples/nuxt-2-composition-api/README.md
```diff
@@ -0,0 +1,68 @@
+# nuxt-starter
+
+## Build Setup
```
mdarrik, line 3: Since this is an example instead of a starter, it might be worth adding a bit more of a description to the Readme. Giving a little background on the example, and what users might want to get from it. Possibly how to clone it as a starter if they want to do so.

alexcasche:
Closing in favor of https://github.com/getnacelle/nacelle-js/pull/65
 - this branch fell out of step with main and it was easier to re-open a fresh PR

PR #73, opened 2022-02-02
fix(starters/gatsby): remove duplicate layout component

alexcasche:
**What This PR Does**
- removes `Layout` component from `wrapRootElement` so it doesn't render twice


NWRichmond (approved):
Great! Thank you @alexcasche.

PR #82, opened 2022-02-23
feat: add fragments for checkouts

alexcasche:
**What this PR does**
- adds new gql fragments for `lineItem`, `discountApplication`, `checkout`, `userError`
- uses fragments to queries and mutations

**Notes**
I think this is what was intended by ticket.  If not happy to make updates

### DIFF
--- a/packages/shopify-checkout/src/graphql/fragments/index.ts
+++ b/packages/shopify-checkout/src/graphql/fragments/index.ts
@@ -1,0 +1,82 @@
+import { gql } from '../../utils';
+
+const lineItem = gql`
+  fragment CheckoutLineItem_lineItem on CheckoutLineItem {
+    customAttributes {
+      key
+      value
+    }
+    discountAllocations {
+      allocatedAmount {
+        amount
+        currencyCode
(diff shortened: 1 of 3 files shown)


NWRichmond:
Hi @alexcasche! The fragments look great.

I think we probably don't want this change on `main` yet - we'll probably want changes to go to `main` after [LS-1548](https://nacelle.atlassian.net/browse/LS-1548) is complete.

I just created a new branch, [`LS-1546/line-item-discounts`](https://github.com/getnacelle/nacelle-js/tree/LS-1546/line-item-discounts), that we can use as a collection point for changes that will lead to the `0.0.9` release. Could you please submit the PR to that branch instead? And could you please also add a link to the ticket ([LS-1547](https://nacelle.atlassian.net/browse/LS-1547)) in the PR description?

jasonredick (commented):
The fragments look good. It seems like the client actions will still only return `id`, `url`, and `completed`. Were these supposed to be updated to account for the new data being requested in these fragments or is that to be a separate task?

PR #84, opened 2022-02-25
feat(shopify-checkout): add in shopify scripts support

alexcasche:

Adds [[LS-1548](https://nacelle.atlassian.net/browse/LS-1548)] 

- Adds support for `discountApplications` and `lineItems` to return all Shopify Scripts related information
- Locks the package to specified version of Shopify Storefront API

**Note:** The approach here was highly dictated by the upcoming rewrite of `shopify-checkout` outlined [here](https://nacelle.atlassian.net/wiki/spaces/DD/pages/1667301474/nacelle+shopify-checkout+version+1.0.0#Implementation-Strategy).

### HOW to test?

`git checkout LS-1548--line-item-mutations-queries`
`npm run test:ci`

### NOTES

Looks like the `next-starter` is failing to build on Vercel.  Will create a separate Jira ticket for this

### DIFF
--- a/packages/shopify-checkout/src/utils/buildCheckout.ts
+++ b/packages/shopify-checkout/src/utils/buildCheckout.ts
@@ -2,11 +2,15 @@ import { BuildCheckoutParams, ShopifyCheckout } from '../checkout-client.types';
 
 export default function buildCheckout({
   id,
-  webUrl
+  webUrl,
+  lineItems,
+  discountApplications
 }: BuildCheckoutParams): ShopifyCheckout {
   return {
     id,
     url: webUrl,
-    completed: false
+    completed: false,
+    lines: lineItems.edges.map((edge) => edge.node),
+    discounts: discountApplications.edges.map((edge) => edge.node)
   };
 }
--- a/packages/shopify-checkout/src/utils/buildCheckout.ts
+++ b/packages/shopify-checkout/src/utils/buildCheckout.ts
@@ -3,13 +3,14 @@ import { BuildCheckoutParams, ShopifyCheckout } from '../checkout-client.types';
 export default function buildCheckout({
   id,
   webUrl,
+  completedAt,
   lineItems,
   discountApplications
 }: BuildCheckoutParams): ShopifyCheckout {
   return {
     id,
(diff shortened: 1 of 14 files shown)


packages/shopify-checkout/src/utils/buildCheckout.ts
```diff
@@ -2,11 +2,16 @@ import { BuildCheckoutParams, ShopifyCheckout } from '../checkout-client.types';
...
+  lineItems,
+  discountApplications
 }: BuildCheckoutParams): ShopifyCheckout {
   return {
     id,
     url: webUrl,
-    completed: false
+    completed: Boolean(completedAt),
+    lines: lineItems.edges.map((edge) => edge.node),
+    discounts: discountApplications.edges.map((edge) => edge.node)
```
jasonredick, line 15: Could we write tests for these lines? This file has pretty low coverage (probably cause it's so small) and running the test points to these two lines.

packages/shopify-checkout/src/utils/createGqlClient.ts
```diff
@@ -41,18 +41,21 @@ export default function createGqlClient({
...
-  storefrontApiVersion,
   storefrontCheckoutToken
 }: CreateGqlClientParams): GqlClient {
   const gqlClient: GqlClient = ({ query, variables }) => {
-    let endpoint = customEndpoint;
+    let endpoint = customEndpoint || '';
+
+    if (!storefrontCheckoutToken) {
+      throw new Error(missingAccessTokenMessage);
+    }
```
jasonredick, line 51: We should probably test for this error too.

jasonredick (approved):
This looks good to me. There are some improvements to be made in our tests, but I don't know if it's enough to hold this up.

NWRichmond (approved):
Thank you for the coverage improvements in https://github.com/getnacelle/nacelle-js/pull/84/commits/487d86b8a0c32743cd2923ac93e9e7f835ff4d26.

In local testing, I found that Shopify Scripts discounts were returned in `cart.lines[x].discountAllocations` and `cart.discounts`. Excellent!

PR #85, opened 2022-02-28
refactor: remove locale env from starters

alexcasche:

Fixes [[LS-1572](https://nacelle.atlassian.net/browse/LS-1572)] 

Removes the `locale` env values passed to `Storefront` sdk in starters.

##


NWRichmond (changes requested):
Thanks for this!

Looks like we'll just also need to remove this line: https://github.com/getnacelle/nacelle-js/blob/LS-1572--remove-default-locale/starters/nuxt/utils/buildRoutes.js#L7

KrisQ (approved):
LGTM!

NWRichmond (commented):
Thanks for that update!

PR #90, opened 2022-03-04
feat(examples/flash-sale): adds flash sale example

jasonredick:

Story [[LS-1560](https://nacelle.atlassian.net/browse/LS-1560)] 

Adds an example store demonstrating a flash sale, with flash content being provided by Contentful.

### How to Test

1. Run `npm run bootstrap` in the root.
2. In `examples/flash-sale`, add a `.env` file. Use the credentials provided below.
3. Run `npm run dev`.
4. Navigate to `/collections/clothing` to see the flash sale timer.

```
# Nacelle
NEXT_PUBLIC_NACELLE_STOREFRONT_ENDPOINT="https://storefront.api.nacelle.com/graphql/v1/spaces/f79296c6-b8dc-4b05-bc46-7305bc8728e3"
NEXT_PUBLIC_NACELLE_STOREFRONT_TOKEN="c5cd9e7c-a049-417b-a5ce-05d575632d2f"

# Shopify Checkout (for details, see: https://www.npmjs.com/package/@nacelle/shopify-checkout)
NEXT_PUBLIC_MYSHOPIFY_DOMAIN="prairie-wind-apparel-sales"
NEXT_PUBLIC_SHOPIFY_STOREFRONT_CHECKOUT_TOKEN="55db5e8a1325c82e275c5fc9301df711"
NEXT_PUBLIC_STOREFRONT_API_VERSION="2022-01"
```


examples/flash-sale/pages/collections/[handle].js
```diff
@@ -0,0 +1,233 @@
...
+  const flashSaleRenderer = (timeDelta) => {
+    const formattedDate = formatTimeDelta(timeDelta, {
+      daysInHours: true
+    });
+    const { hours, minutes, seconds, completed } = formattedDate;
+    if (completed) {
+      return <></>;
+    } else {
+      return (
+        <p suppressHydrationWarning>
```
Comment (alexcasche) [inline examples/flash-sale/pages/collections/[handle].js:29]: More of a question than a request.  Is `suppressHydrationWarning` a good approach?  How does it stack up to conditional rendering like `typeof window !== 'undefined' { // client side code }`

alexcasche (commented):
Looks good to me.  Two things I'm hoping others can weigh in on before we approve.

1.  Should we make the example project name more specific?  Ex: `nextjs-flash-sale`
2.  Should we keep the sale content locally instead of linking to Contentful?  Benefit there is we could always set the flash sale for say "5 minutes > now"

examples/flash-sale/pages/collections/[handle].js
Comment (jasonredick) [reply examples/flash-sale/pages/collections/[handle].js:29]: I added it to get rid of a console error: https://github.com/ndresx/react-countdown#why-do-i-get-this-error-warning-text-content-did-not-match

jasonredick:
> Looks good to me. Two things I'm hoping others can weigh in on before we approve.
> 
> 1. Should we make the example project name more specific?  Ex: `nextjs-flash-sale`
> 2. Should we keep the sale content locally instead of linking to Contentful?  Benefit there is we could always set the flash sale for say "5 minutes > now"

1. That sounds good to me if that's the convention we want to use.
2. I originally had the sale content hardcoded into the collection template and left the query as a comment. I was concerned about it not being a true representation of the query working though. I could add it back, as then we wouldn't have to worry about that content becoming stale.

alexcasche:
> > Looks good to me. Two things I'm hoping others can weigh in on before we approve.
> > 
> > 1. Should we make the example project name more specific?  Ex: `nextjs-flash-sale`
> > 2. Should we keep the sale content locally instead of linking to Contentful?  Benefit there is we could always set the flash sale for say "5 minutes > now"
> 
> 1. That sounds good to me if that's the convention we want to use.
> 2. I originally had the sale content hardcoded into the collection template and left the query as a comment. I was concerned about it not being a true representation of the query working though. I could add it back, as then we wouldn't have to worry about that content becoming stale.

I'm good either way.  Let's see what the other Lyrans have to say

examples/flash-sale/pages/collections/[handle].js
Comment (NWRichmond) [inline examples/flash-sale/pages/collections/[handle].js:29]: To Alex's point about rendering only on the client, one way we could accomplish that would be to leverage Webpack's `process.browser`:

```suggestion
    if (completed || !process.browser) {
      return <></>;
    } else {
      return (
        <p>
```

IMHO that'd be preferable to suppressing hydration warnings.

NWRichmond (changes requested):
Excellent! I'm digging this, and am particularly fond of the README, which strikes a nice balance of detail & brevity.

In addition to the suggestion about hydration, there are a couple of things that I think will help give this project some added staying power:

1. **Timer** - I think there's an opportunity to keep the content as the primary source of truth for the sale timing, but if it’s expired, fall back to a time shortly in the future, with code comments to explain that the fallback code is for demonstration purposes only. This will allow us to deploy this project to Vercel and have it be more meaningful to onlookers without needing to make changes in the CMS.
2. **Framework Specificity** - To address @alexcasche's [question](https://github.com/getnacelle/nacelle-js/pull/90#pullrequestreview-900643983):
> Should we make the example project name more specific? Ex: nextjs-flash-sale

I think it'd be great if we could keep this as generic as possible, so that we don't have people asking us, "When are you going to make a flash sale example in [my framework of choice]?" So I think the name of this project is good. But I think there might be some opportunities in the README to call out things like (maybe more formally than I'm about to say), "This project happens to use React and Next.js, but the concepts are applicable to any JavaScript project. If you're using Vue, consider using [some countdown package]." That way, this project's focus can be on the concepts & data patterns, rather than the component framework.

examples/flash-sale/pages/collections/[handle].js
Comment (alexcasche) [reply examples/flash-sale/pages/collections/[handle].js:29]: > To Alex's point about rendering only on the client, one way we could accomplish that would be to leverage Webpack's `process.browser`:
> 
> IMHO that'd be preferable to suppressing hydration warnings.

Should we use `typeof window !== undefined`?  Seems like the `process.browser` is being deprecated:
https://github.com/vercel/next.js/issues/5354#issuecomment-520305040

Comment (NWRichmond) [reply examples/flash-sale/pages/collections/[handle].js:29]: Yes, good catch @alexcasche! I've updated the suggestion accordingly.

Comment (jasonredick) [reply examples/flash-sale/pages/collections/[handle].js:29]: While this does work, it leaves me with a new error in the console.

Screen Shot 2022-03-09 at 9 12 24 AM

Comment (jasonredick) [reply examples/flash-sale/pages/collections/[handle].js:29]: @alexcasche @NWRichmond I assume what's happening here is that it renders on the server with no elements, but on the client it has the `<p>` and `<strong>` tag from the countdown timer markup.

The use of `suppressHydrationWarning` was to get around the timer running and being different between server and client. I've tried a different approach, having the timer not automatically start, and start it manually, but was running into the same warning error of the timer being different.

Comment (alexcasche) [reply examples/flash-sale/pages/collections/[handle].js:29]: Got a new solution @jasonredick @NWRichmond .  Kind of annoying but seems like the Next.js way.

```
// Step 1:  Add new piece of state
const [showCountdown, setShowCountdown] = useState();
```

```
// Step 2:  Toggle to true in useEffect
useEffect(() => {
    setShowCountdown(true)
  }, [])
```

```
// Step 3:  Use this variable to conditionally render content
if (completed || !showCountdown) {
      return <></>;
    } else {
      return (
        <p>
          <strong>Sale Ends</strong> {zeroPad(hours)}:{zeroPad(minutes)}:
          {zeroPad(seconds)}
        </p>
      );
    }
```

Comment (jasonredick) [reply examples/flash-sale/pages/collections/[handle].js]: @alexcasche That looks like it worked, thank you!

alexcasche (approved):
Looks good!

NWRichmond (approved):
Right on! Thanks for those updates, @jasonredick!

After this is merged, I think it'd be helpful to Code Sandbox and add the link to the README.

PR #92, opened 2022-03-09
add: vue-shopify-account example

KrisQ:
**Ticket**

https://nacelle.atlassian.net/browse/LS-1553

**Changes**

Very minimal app.
Add an example Vue project for Shopify accounts:
- login
- register
- recover
- reset
- customer details
- addresses
- orders
- multipass login for checkout

**Testing**

- Make sure you have the [vercel CLI](https://vercel.com/cli) installed
- Reach out to me for env if you'd like, as it contains admin password (multipass)
- `npm i dev:vercel`
- When prompted there should already be a vercel project `vue-shopify-account`
- I have been using a [temp email](https://temp-mail.org/en/) for my testing and found it very useful


examples/vue-shopify-account/README.md
```diff
@@ -0,0 +1,142 @@
...
+3. you will also need the [vercel CLI](https://vercel.com/cli) installed in order to test serverless functions
+
+## Serverless Functions
+
+To use Multipass and the address form for account pages this project relies on two packages:
+
+- [multipassify](https://github.com/beaucoo/multipassify)
+- [countrycitystatejson](https://github.com/khkwan0/countryCityStateJson)
+
+Both of these packages are _large_ and can add a lot to your client bundle, so using them only in serverless functions keeps the client a little more lightweight. But if you want another reason, it also keeps your Multipass secret outside of client code.
```
alexcasche, line 44: Nice call out about keeping the Multipass secret outside the client!

examples/vue-shopify-account/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+This project includes a folders for serverless functions:
```
alexcasche, line 46:
```suggestion
This project includes a folder for serverless functions:
```

examples/vue-shopify-account/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+- Vercel in [api/\*](./api)
+
+When testing your serverless functions locally make sure to use the platform's CLI ([Vercel CLI](https://vercel.com/docs/cli). NPM scripts for running the project with these CLI's are provided in the `package.json`.
+
+## Functions
+
+All the functions to interact with the API live in the `/components/AccountProvider.vue` file and the graphql queries are in the `/gql/index.js`.
+
+We are using [the composition API](https://vuejs.org/guide/extras/composition-api-faq.html) for the state management but this can be integrated into any state management system.
```
alexcasche, line 56:
```suggestion
We are using [the composition API](https://vuejs.org/guide/extras/composition-api-faq.html) for state management but this can be integrated into any state management system.
```

examples/vue-shopify-account/README.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Multipass Login
+
+In the `/component/AccountProvider.vue` file there is a `multipassLogin` method, this will log in the user in Shopify. The best use for this function is to login the user before checkout, carrying over their information to the checkout page.
+
+## Using Nuxt?
+
+If you implementing this system into a nuxt project we recommend making use of [middleware](https://nuxtjs.org/docs/directory-structure/middleware/) to protect authenticated routes:
```
alexcasche, line 64:
```suggestion
If you are implementing this system into a nuxt project we recommend using [middleware](https://nuxtjs.org/docs/directory-structure/middleware/) to protect authenticated routes:
```

examples/vue-shopify-account/README.md
```diff
@@ -0,0 +1,142 @@
...
+  // If the user is authenticated redirect to account
+  const customerAccessToken = app.$cookies.get('customerAccessToken')
+
+  if (customerAccessToken) {
+    return redirect('/account')
+  }
+}
+```
+
+These example assume the adition of the `cookie-universal-nuxt` [package](https://www.npmjs.com/package/cookie-universal-nuxt).
```
alexcasche, line 94:
```suggestion
These examples assume the addition of `cookie-universal-nuxt` [package](https://www.npmjs.com/package/cookie-universal-nuxt).
```

examples/vue-shopify-account/api/customer-transactions.js
```diff
@@ -0,0 +1,24 @@
+// Expected request body:
+//
+//  { orderID: "the-order-id" }
```
alexcasche, line 3: Good call including these example payloads!

examples/vue-shopify-account/src/components/Account.vue
```diff
@@ -0,0 +1,79 @@
...
+    })
+
+    onMounted(async () => {
+      await stayLoggedIn()
+      window.addEventListener('hashchange', () => {
+        currentPath.value = window.location.hash
+      })
+    })
+    return {
+      currentPath,
```
alexcasche, line 54:
```suggestion
```

alexcasche: I don't think we need to return this

alexcasche (changes requested):
This is really nice Kristenn!  You definitely put a lot of work in here.  Much appreciated!

alexcasche (changes requested):
One other thing, which is definitely my fault.  I guess we've decided to not framework namespace projects.  So many this is just called `examples/shopify-multipass-accounts`

Hey @alexcasche I think considering the scope of this and it doing more than multipass, should we go with simply:

`shopify-accounts`

?

alexcasche:
@KrisQ the project works well for me.  As a dev I think it does what it needs to.  The UI is barebones, without loading states, but I think that is the point.  Open to what others have to say, but to me this project serves its purpose nicely

alexcasche:
> > One other thing, which is definitely my fault. I guess we've decided to not framework namespace projects. So many this is just called `examples/shopify-multipass-accounts`
> 
> Hey @alexcasche I think considering the scope of this and it doing more than multipass, should we go with simply:
> 
> `shopify-accounts`
> 
> ?

I'm good with that.  In the future, we may want to separate multipass from the 'addresses, orders, etc' functionality.  Those are really storefront API integrations.  But I think this is perfect for now.

NWRichmond (changes requested):
@KrisQ thanks for this PR!

Because the scope of [LS-1553](https://nacelle.atlassian.net/browse/LS-1553) is email login only, we don't need to use Multipass. Removing Multipass will help to reduce complexity and extra steps for the merchant developer. We have another story, [LS-1554](https://nacelle.atlassian.net/browse/LS-1554), that does require use of Multipass to power a social auth experience.

KrisQ:
@alexcasche and @NWRichmond thank you for all your reviews, everything should be fixed :)

alexcasche (approved):

PR #103, opened 2022-03-15
fix(shopify-checkout): accept checkout IDs that start with letter

NWRichmond:

Fixes [ENG-4354](https://nacelle.atlassian.net/browse/ENG-4354)

> @nacelle/shopify-checkout is erroneously reporting checkout ids as invalid

Our validation of Shopify checkout IDs in `packages/shopify-checkout/src/utils/isVerifiedCheckoutId.ts` was not accounting for an aspect of Base64 encoding.

Notice how in the following example, when the Checkout ID starts with a letter, the Base64-encoded string has a trailing `1dC9`, whereas the absence of a Checkout ID or a Checkout ID that starts with a number has a trailing `1dC8`:

base64 encoding shopify checkout example

Because we used `'Z2lkOi8vc2hvcGlmeS9DaGVja291dC8'` as a comparison string, the `isVerifiedCheckoutId` utility was incorrectly returning `false` when a Checkout ID began with a letter.

This PR changes `isVerifiedCheckoutId`'s comparison string to `'Z2lkOi8vc2hvcGlmeS9DaGVja291dC'` (it removes the trailing `8`), so that Checkout IDs that start with either letters or numbers are returned as `true`. It also add and updates the appropriate tests to account for both cases.

## 

### How to Test

- Check out the `fix/ENG-4354--nacelle-shopify-checkout-isVerifiedCheckoutId` branch
- `cd packages/shopify-checkout`
- `npm run test:ci` (all tests should pass)
nacelle-js fix:ENG-4354--nacelle-shopify-checkout-isVerifiedCheckoutId all tests passing

- `npm run build` (it should build all outputs without errors)
nacelle-js fix:ENG-4354--nacelle-shopify-checkout-isVerifiedCheckoutId build outputs

### DIFF
--- a/packages/shopify-checkout/__tests__/mocks/index.ts
+++ b/packages/shopify-checkout/__tests__/mocks/index.ts
@@ -17,14 +17,22 @@ export const clientSettings = {
 
 export const graphqlEndpoint = `https://${clientSettings.myshopifyDomain}.myshopify.com/api/2022-01/graphql`;
 
-const checkoutUuidWithKey = '998877?key=123123';
-export const checkoutId = Buffer.from(
-  'gid://shopify/Checkout/' + checkoutUuidWithKey
-).toString('base64');
+const checkoutUuids = {
+  beginsWithLetter: 'a9b8c7?key=123123',
+  beginsWithNumber: '998877?key=123123'
+};
+export const checkoutIds = {
+  beginsWithLetter: Buffer.from(
+    'gid://shopify/Checkout/' + checkoutUuids.beginsWithLetter
+  ).toString('base64'),
+  beginsWithNumber: Buffer.from(
+    'gid://shopify/Checkout/' + checkoutUuids.beginsWithNumber
+  ).toString('base64')
+};
 
 export const webUrl =
   'https://nacelle-swag-store.myshopify.com/112233/checkouts/' +
-  checkoutUuidWithKey;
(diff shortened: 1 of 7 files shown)


jasonredick (approved):
Looks good to me. Tests are passing and successfully builds.

mdarrik (approved):
Not sure if we want to consider unifing this with what we do in [transformVariantIdToGid.ts](https://github.com/getnacelle/nacelle-js/blob/4df7224d0dd6056717eb48be6f81278e3b00b759/packages/shopify-checkout/src/utils/transformVariantIdToGid.ts#L18) 
Where we decode & then check the url or not. But for fixing the existing bug this seems good,  and all tests pass.

NWRichmond:
> Not sure if we want to consider unifing this with what we do in [transformVariantIdToGid.ts](https://github.com/getnacelle/nacelle-js/blob/4df7224d0dd6056717eb48be6f81278e3b00b759/packages/shopify-checkout/src/utils/transformVariantIdToGid.ts#L18) Where we decode & then check the url or not. But for fixing the existing bug this seems good, and all tests pass.

I think that'd be a great follow-up PR!

Thanks for the review 🤝

NWRichmond:
Published: https://www.npmjs.com/package/@nacelle/shopify-checkout/v/0.0.10

PR #120, opened 2022-04-26
feat: Algolia example

levimichael:
Added an Algolia usage example which contains a boilerplate indexing service (Nacelle -> Algolia) and a Stackblitz link.


examples/with-algolia/service/README.md
```diff
@@ -0,0 +1,26 @@
...
+TODO
+
+## Tools Used
+
+- AWS Lambda
+- Severless Framework
+- Typescript (via `serverless-plugin-typescript`)
+- Express.js
+- Jest
+- DataDog
```
NWRichmond, line 26: I recommend updating this README to include context & instructions for the Merchant Developer audience.

examples/with-algolia/service/serverless.yml
```diff
@@ -0,0 +1,41 @@
...
+#     logLevel: "info"
+#     enableXrayTracing: false
+#     enableDDTracing: true
+#     enableAPIGatewayLogs: true
+#     forwarderArn: arn:aws:lambda:us-east-1:000000000000:function:datadog-forwarder
+#     enableTags: true
+#     injectLogContext: true
+#     exclude:
+#       - dd-excluded-function
+# info https://docs.datadoghq.com/serverless/installation/nodejs/?tab=serverlessframework
```
NWRichmond, line 41: Can the commented-out parts be removed?

NWRichmond (changes requested):
This looks good! Just have a couple of recommendations:

examples/with-algolia/service/README.md
levimichael: 🤦‍♂️ yep. very good point. :D will fix.

examples/with-algolia/service/serverless.yml
levimichael: Yes, will do!

levimichael:
@NWRichmond I removed the unnecessary readme and updated the serverless.yml.

examples/with-algolia/service/serverless.yml
```diff
@@ -0,0 +1,22 @@
...
+functions:
+  api:
+    handler: handler.handler
+    events:
+      - httpApi: '*'
+    environment:
+      ALGOLIA_INDEX: ${env:ALGOLIA_INDEX}
+      ALGOLIA_API_KEY: ${env:ALGOLIA_API_KEY}
+      STOREFRONT_API_ENDPOINT: ${env:STOREFRONT_API_ENDPOINT}
+      STOREFRONT_API_TOKEN: ${env:STOREFRONT_API_TOKEN}
```
NWRichmond, line 22: Given potential for confusion with the Shopify Storefront API, probably worth prefixing these env vars with `NACELLE_`.

NWRichmond (changes requested):
Thanks for those updates, @levimichael!

Sorry for not catching this in the earlier review, but I think some more detail / framing would be useful in `examples/with-algolia/README.md`. Here are some things that, as a merchant developer, I'd be looking for in that README:
1. What is Algolia?
2. How does this relate to Nacelle? (Looking at the code, it looks like we're pulling data from Nacelle and pushing it to Algolia. But it would be great to have high-level concepts like that spelled out in the README).
3. What is the expected result of deploying the service? In other words, "how can I use this once it's deployed?"
4. What does someone need to do to deploy this code? For example, [some environment variables are needed](https://github.com/getnacelle/nacelle-js/blob/09c48f9a005b7d6bc131954d890cec5402a3ebf1/examples/with-algolia/service/serverless.yml#L19-L22). But some explanation about where to find those variables & where to put them (`.env`? Somewhere in AWS?) would be very helpful.

As far as Example Project READMEs go, [`examples/flash-sale`](https://github.com/getnacelle/nacelle-js/tree/main/examples/flash-sale#flash-sale) and [`examples/shopify-accounts`](https://github.com/getnacelle/nacelle-js/tree/main/examples/shopify-accounts#accounts-example) are probably good reference points. I think it'd be okay to have `examples/with-algolia`'s README be not-quite-as-verbose, but there's probably a nice balance that can be struck here.

examples/with-algolia/service/serverless.yml
levimichael: Updated!

levimichael:
@NWRichmond ready for your review again! Thanks for the good callouts.

levimichael:
Fantastic, thanks @NWRichmond

PR #121, opened 2022-04-26
feat(gatsby-source-nacelle): handle nested content references

mdarrik:

Addresses [ENG-4700](https://nacelle.atlassian.net/browse/ENG-4700) 

- Replaces nested content nodes with links so Gatsby can link them to their top level nodes. This also allows arrays of content references to have different fields, which was broken before. 

- During source nodes, searches for arrays/objects inside the `remoteFields` key of content & replaces any object with a `nacelleEntryId` with a `___NODE` reference to the content node with the same Gatsby Id. For arrays, replaces the entire array with an array of Gatsby Node Id's. This allows for users to query their inferred, space-specific NacelleContent types using interface fragments. 

### Assumptions/Caveats

- Currently treats all objects with `nacelleEntryId` keys as if they are Content references.
- For custom ingestion/other sources, this won't handle arrays of disparate types that don't have nacelleEntryId's. As long as they don't have field name overlap with different types, Gatsby's native inference should handle those cases okay.  

## How to Test

1. Navigate to `packages/gatsby-source-nacelle` in your terminal
2. Run `npm link`
3. Navigate to either `starters/gatsby` or an existing/newly created Nacelle Gatsby Repo
   - Note that it's helpful to use a space with more complex content models that has internal content links.
4. Run `npm link @nacelle/gatsby-source-nacelle` to link to the version in this PR
5. Run `npm run develop` to start the dev server
6. Navigate to http://localhost:8000/___graphql and explore the queries for various content types
   - For arrays of content, should be able to query each of the content types included in the array using `...on NacelleContentRemote<TypeName>` fragment queries.


NWRichmond (commented):
Excellent work, @mdarrik! I really appreciate the helpful code comments.

In testing, arrays of content with heterogeneous `type`s were assigned a [union type](https://graphql.org/learn/schema/#union-types) with all of the possible & actual types:

gatsby-array-of-content-multiple-types

gatsby-query-array-multiple-content-types

I'm thrilled to see that it didn't take many new lines of code to get there 👏

I think this is going to be a great approach. Pending some updates to the `README` to provide a brief overview of the query syntax, I think this is good to go!

packages/gatsby-source-nacelle/src/source-nodes/index.js
```diff
@@ -82,10 +83,11 @@ module.exports = async function ({
...
                 ])?.filter((val) => val);
-                // create remote image file nodes for each field
-                await Promise.all(
-                  fetchRemoteImageNodesForContent(node, images, gatsbyApi)
-                );
+                await findNestedReferences(node.remoteFields, node),
+                  // create remote image file nodes for each field
+                  await Promise.all(
+                    fetchRemoteImageNodesForContent(node, images, gatsbyApi)
+                  );
```
NWRichmond, line 90: Looks like there was a suspect `,`:

```suggestion
                await findNestedReferences(node.remoteFields, node)
                // create remote image file nodes for each field
                await Promise.all(
                  fetchRemoteImageNodesForContent(node, images, gatsbyApi)
                );
```

packages/gatsby-source-nacelle/src/source-nodes/index.js
```diff
@@ -6,6 +6,7 @@ const {
   replaceKey,
   findNestedImages
 } = require('../utils');
+const findNestedReferences = require('../utils/linkNestedContentReferences');
```
NWRichmond, line 9: Could we add this to `packages/gatsby-source-nacelle/src/utils/index.js` and import it alongside the other utils?

jasonredick (commented):
Everything looks good to me. No issues setting it up and testing. I think we're good once Nick's comments are resolved.

packages/gatsby-source-nacelle/src/source-nodes/index.js
mdarrik: This should be fixed now! Thanks for the catch!

packages/gatsby-source-nacelle/src/source-nodes/index.js
mdarrik: Yeah. Done!

mdarrik:
merged main into gatsby-source-nacelle-9.0 branch and then that branch into this one

NWRichmond (approved):
Thanks for merging the latest changes from [`gatsby-source-nacelle-9.0`](https://github.com/getnacelle/nacelle-js/tree/gatsby-source-nacelle-9.0)! Great work here @mdarrik.

PR #136, opened 2022-05-06
chore: close search autocomplete on enter

alexcasche:
**What This PR Does**
Ensures the search autocomplete dropdown closes on enter keypress.  

**How To Test**
Spin up this branch.  Enter some search terms in search input and hit enter.  The autocomplete dropdown should close

### DIFF
--- a/reference-stores/nuxt/components/search/SearchInput.vue
+++ b/reference-stores/nuxt/components/search/SearchInput.vue
@@ -110,6 +110,7 @@ export default {
       });
     },
     handleEnter(event) {
+      this.isFocussed = false;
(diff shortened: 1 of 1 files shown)


alexcasche:
> Looks good.
> 
> Small accessibility nit which is likely more for documentation - if we're blurring focus off the input on navigation, we need to make sure it's set somewhere. (Probably need to handle focus for the store as a whole on navigation, but this one stuck out because we manually blur focus)

Totally.  I think we'll make a number of good accessibility improvements when we do the full sweep.  Anything we find in Nuxt will also be in Next/Gatsby so hitting it in one sweep will be great.

PR #138, opened 2022-05-09
feat(gatsby)!: infer types for `Content.fields`

NWRichmond:
### Ticket

[ENG-4001](https://nacelle.atlassian.net/browse/ENG-4001)

### Problem statement

[`@nacelle/gatsby-source-nacelle`](https://www.gatsbyjs.com/plugins/@nacelle/gatsby-source-nacelle) version `8.x` could not source `remoteImage` data from Nacelle v2 `Content` due to image fields being held in sub-properties of `Content.fields`, which is presently typed as `JSON` in the Nacelle Storefront GraphQL schema. Because those fields were typed as `JSON`, Gatsby didn't have enough knowledge of the `Content` data to be able to create [image nodes](https://www.gatsbyjs.com/docs/how-to/images-and-media/preprocessing-external-images) for [`gatsby-plugin-image`](https://www.gatsbyjs.com/plugins/gatsby-plugin-image) at arbitrary data depths.

### Main thrust

This PR is the culmination of #104 and #121. By leveraging [Gatsby schema inference](https://www.gatsbyjs.com/docs/schema-inference), we add rich type information to content data. Some benefits and implications of this:

#### Images

Gatsby image nodes can now be sourced at arbitrary depths in content queries.

#### Querying content data

Rich type information enables merchant developers to make granular queries for the fields of interest. This is likely to offer a significant reduction in payload sizes for content queries. Implications include:
1. Instead of one `NacelleContent` type, types are now created for each content model. These content types are bespoke to a Nacelle space.
2. Merchant developers have access to new root-level queries corresponding to their bespoke content types.
3. When querying arrays of content data that contain multiple content types, [inline fragments](https://graphql.org/learn/queries/#inline-fragments) must be used to select fields for each type.

### Supporting & miscellaneous changes

1. In the `README`s for `starters/gatsby` and `packages/gatsby-source-nacelle`, add links to Gatsby docs which explain the new query patterns, including inline fragments. The docs updates will be published after this PR is merged, but they can be previewed [here](https://nacelle.atlassian.net/wiki/spaces/PT/pages/1613627437/Guide+Using+Gatsby#Data-Fetching).
4. Updates `packages/gatsby-source-nacelle/LICENSE` to use the `nacelle-js` Apache-2.0 license.

## 

### How to Test

Many of the instructions below are borrowed from #104.

1. Check out the `ENG-4001/gatsby-source-nacelle-9.0` branch.
4. `npm run bootstrap` in the monorepo root if you haven't already.
[...]

PR #143, opened 2022-05-12
build(deps): manually bump deps

alexcasche:
## What This PR Does

Manually bumps dependencies in all projects to get broken dependabot PRs closed.

## Caveats

This does not address security vulnerabilities in packages.  To accommodate that we would need to run `npm audit fix --force` and manually address any breaking changes.  If that is desired we can address it in another PR.


NWRichmond (approved):
Good stuff! Thank you for excluding `starters/gatsby`, which will be taken care of by `https://github.com/getnacelle/nacelle-js/pull/138`.

PR #145, opened 2022-05-26
feat(gatsby): scaffold gatsby reference store 

alexcasche:

Addresses 
 - [ENG-4485](https://nacelle.atlassian.net/browse/ENG-4485) 
 - [ENG-5482](https://nacelle.atlassian.net/browse/ENG-5482) 

- Scaffolds a Gatsby.js reference store using tailwind, and Contentful.

- Scaffolds Gatsby reference store and adds layout component.  Adds minor patches to Next.js reference store.

## 

### Gotchas

- Due to `gatsby-source-nacelle` M1 mac issues, this does not utilize `gatsby-source-filesystem`.  In order to make it work you need to ensure that `useGatsbyImage` is false in `gatsby-source-nacelle`.  To do so you should comment out this line of code in `node_modules/@nacelle/gatsby-source-nacelle/src/source-nodes/index.js` 
```
try {
    // require('gatsby-source-filesystem');
    // useGatsbyImage = true;
}
```
- Since we are not currently using `gatsby-image` some images on the site may look misformatted.  This will be addressed in a follow-up PR once we have figured out the M1 mac issues.

### Testing
- `git checkout ENG-5482/gatsby-reference-store`
- `cd references-stores/gatsby && npm install`
- `npm run develop`
- Jump around the site and see that the layout components render as expected (`Header`, `Footer`, `Nav`, `Cart`, etc.


reference-stores/gatsby/src/services/nacelleClient.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default {
+
+}
+
+// export default new Storefront({
+//   storefrontEndpoint: process.env.GATSBY_PUBLIC_NACELLE_STOREFRONT_ENDPOINT,
+//   token: process.env.GATSBY_PUBLIC_NACELLE_STOREFRONT_TOKEN,
+//   locale: process.env.GATSBY_PUBLIC_NACELLE_STOREFRONT_LOCALE
+// });
```
jasonredick, line 13: Don't believe this file is necessary.

jasonredick (approved):
Great work! Was able to get Gatsby up and running after making the change mentioned in the instructions. I mention a small change below, but otherwise it looks good to me.

PR #174, opened 2022-07-08
feat: shopify checkout custom query

KrisQ:

Fixes [[ENG-6125](https://nacelle.atlassian.net/browse/ENG-6125)] 

- add Shopify checkout custom query method
- update Shopify API version to 2022-07

## 

### How to Test

🖥 [Local development instructions](#link-to-repo-readme)
🗒 [General testing guidelines](#link-to-testing-guidelines)
📄 [Changelog guidelines](#link-to-changelog-guidelines)

- `npm run test` everything should pass

For end to end testing
- there is a branch of the [shopify-checkout-react-discounts-testing](https://github.com/getnacelle/shopify-checkout-react-discounts-testing/tree/ENG-6125--shopify-custom-query) that is made to test this feature
- for me `npm run build && npm run preview` worked better than `npm run dev` for some reason
- click "Process Checkout" and look at the request in the network tab, make sure that the endpoint uses the `2022-07` version rather than the `2022-01`
- click on the "Custom Query" button and look in the inspector, make sure that the object returned is the appropriate object
- navigate to the returned checkout `webUrl` and make sure that the checkout has been updated appropriately

### Testing checklist

* [ ] Tested on mobile
* [ ] Tested on multiple browsers
* [ ] Tested for accessibility
* [x] Updated the `README.md` with documentation changes
* [ ] Updated public facing documentation

### DIFF
--- a/packages/shopify-checkout/README.md
+++ b/packages/shopify-checkout/README.md
@@ -188,17 +188,17 @@ const checkout = await checkoutClient.discountRemove({
 });
 ```
 
-##### Making custom query to Shopify Checkout API
+##### Making custom checkout queries to the Shopify Storefront API
 
-| Parameter | [Type](#common-types)     | Required? | Description                                               |
-| --------- | ------------------------- | --------- | --------------------------------------------------------- |
-| `query`   | `string`                  | ✅        | A valid Shopify API query or mutation                     |
-| `input`   | `Record<string, unknown>` | ✅        | A valid variable objects for a Shopify query or mutations |
+| Parameter   | [Type](#common-types) | Required? | Description                                           |
+| ----------- | --------------------- | --------- | ----------------------------------------------------- |
+| `query`     | `string`              | ✅        | A Shopify Storefront API query/mutation               |
+| `variables` | `object`              | ✅        | Variables for a Shopify Storefront API query/mutation |
 
 ##### Example
 
 ```js
-// Allows for end user to send any query/mutation to Shopify API
+// Send any query/mutation to the Shopify Storefront API
 const checkout = await checkoutClient.query({
   query: `
     mutation checkoutShippingAddressUpdateV2($checkoutId: ID!, $shippingAddress: MailingAddressInput!) {
@@ -215,7 +215,7 @@ const checkout = await checkoutClient.query({
       }
     }
     `,
-  input: {
+  variables: {
     checkoutId: checkoutData.id,
     shippingAddress: {
       address1: '123 smith street',
@@ -230,8 +230,8 @@ const checkout = await checkoutClient.query({
 
 Other example mutations from Shopify docs:
 
-- [checkoutShippingAddressUpdateV2](https://shopify.dev/api/storefront/2022-07/mutations/checkoutShippingAddressUpdateV2)
-- [checkoutCustomerAssociateV2](https://shopify.dev/api/storefront/2022-07/mutations/checkoutCustomerAssociateV2)
+[checkoutshippingaddressupdatev2]: https://shopify.dev/api/storefront/2022-07/mutations/checkoutShippingAddressUpdateV2
+[checkoutcustomerassociatev2]: https://shopify.dev/api/storefront/2022-07/mutations/checkoutCustomerAssociateV2
 
 ### Advanced Usage
 
--- a/packages/shopify-checkout/README.md
+++ b/packages/shopify-checkout/README.md
@@ -188,6 +188,51 @@ const checkout = await checkoutClient.discountRemove({
 });
 ```
 
+##### Making custom query to Shopify Checkout API
+
+| Parameter | [Type](#common-types)     | Required? | Description                                               |
+| --------- | ------------------------- | --------- | --------------------------------------------------------- |
+| `query`   | `string`                  | ✅        | A valid Shopify API query or mutation                     |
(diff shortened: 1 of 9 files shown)


packages/shopify-checkout/src/client/actions/queryCheckout.ts
```diff
@@ -0,0 +1,36 @@
...
+      variables
+    }).catch((err) => {
+      throw new Error(err);
+    });
+
+    return shopifyResponse;
+  } catch (err) {
+    throw new Error(String(err));
+  }
+}
```
NWRichmond, line 36: 1. Because we're not doing anything with the query response (e.g. checking it for `checkoutUserErrors` etc.), we can remove the `try/catch` and simplify this function.
2. We can simplify it further by leaving error handling entirely up to the merchant developer, which IMHO makes sense for "custom queries/mutations". This should help with the test coverage drop as well.
3. For arbitrary queries/mutations, we can't assume that `queueToken` will be a valid variable.

```suggestion
export interface CheckoutQueryParams {
  gqlClient: GqlClient;
  query: string;
  variables: Record<string, unknown>;
}

export default function queryCheckout({
  gqlClient,
  query,
  variables
}: CheckoutQueryParams): Promise<ShopifyResponse<
  Record<string, unknown>
> | void> {
  return gqlClient<typeof variables, ShopifyResponse<Record<string, unknown>>>({
    query,
    variables
  });
}
```

packages/shopify-checkout/src/client/index.ts
```diff
@@ -134,10 +151,19 @@ export default function createShopifyCheckoutClient({
     return await removeDiscount({ gqlClient, id, queueToken });
   }
 
+  async function query({
+    query,
+    input,
+    queueToken
+  }: CustomQueryParams): Promise<ShopifyResponse<unknown> | void> {
+    return await queryCheckout({ gqlClient, query, input, queueToken });
+  }
```
NWRichmond, line 160: An extension of the feedback in `packages/shopify-checkout/src/client/actions/queryCheckout.ts`:

```suggestion
  async function query({
    query,
    variables
  }: CustomQueryParams): Promise<ShopifyResponse<
    Record<string, unknown>
  > | void> {
    return await queryCheckout({ gqlClient, query, variables });
  }
```

NWRichmond (changes requested):
Per the story's **Requirements**, I think we should use the term `variables` instead of `input`.

ENG-6125 query variables

Why? In GraphQL, [`variables`](https://graphql.org/learn/queries/#variables) and [`input`](https://graphql.org/learn/schema/#input-types) are separate concepts. And with the Shopify Storefront API, some mutations do use Inputs (for example, [`checkoutAttributesUpdateV2`](https://shopify.dev/api/storefront/2022-07/mutations/checkoutAttributesUpdateV2). For this reason, `input` could be a confusing or misleading name.

A number of suggestions below assume that we have made the change from `input` to `variables`. There are other suggestions related to error handling and documentation.

KrisQ:
Thank you @NWRichmond for the thorough review, all your suggestions were implemented to this branch!

packages/shopify-checkout/README.md
```diff
@@ -188,6 +188,51 @@ const checkout = await checkoutClient.discountRemove({
...
+      lastName: 'John'
+    }
+  }
+});
+```
+
+Other example mutations from Shopify docs:
+
+[checkoutshippingaddressupdatev2]: https://shopify.dev/api/storefront/2022-07/mutations/checkoutShippingAddressUpdateV2
+[checkoutcustomerassociatev2]: https://shopify.dev/api/storefront/2022-07/mutations/checkoutCustomerAssociateV2
```
NWRichmond, line 234: @KrisQ not quite...these links do not show up:

Screen Shot 2022-07-12 at 12 54 42 PM

Instead:
1. Put these at the bottom of the document, with the other reference-style links (I'd recommend using snake casing to separate the words):

```md
[shopify-mutation-checkout-shipping-address-update-v2]: https://shopify.dev/api/storefront/2022-07/mutations/checkoutShippingAddressUpdateV2
[shopify-mutation-checkout-customer-associate-v2]: https://shopify.dev/api/storefront/2022-07/mutations/checkoutCustomerAssociateV2
```

3. And then here, you can:

```suggestion
- [`checkoutShippingAddressUpdateV2`][shopify-mutation-checkout-shipping-address-update-v2]
- [`checkoutCustomerAssociateV2`][shopify-mutation-checkout-customer-associate-v2]
```

alexcasche (approved):
Minus the final reference-style links comment Nick dropped I think this is good to go

packages/shopify-checkout/src/client/actions/queryCheckout.ts
```diff
@@ -0,0 +1,19 @@
...
+  query: string;
+  variables: Record<string, unknown>;
+}
+export default function queryCheckout({
+  gqlClient,
+  query,
+  variables
+}: CheckoutQueryParams): Promise<ShopifyResponse<
+  Record<string, unknown>
+> | void> {
```
NWRichmond, line 14: Not a big deal, but since we aren't throwing any errors in `queryCheckout` or doing anything else that would prevent us from returning:

```ts
Promise<ShopifyResponse<Record<string, unknown>>>
```
, we could get rid of the ` | void` in the return type:

```suggestion
}: CheckoutQueryParams): Promise<ShopifyResponse<Record<string, unknown>>> {
```

PR #184, opened 2022-07-19
chore: lerna build --no-private

alexcasche:

Fixes [[ENG-6318](https://nacelle.atlassian.net/browse/ENG-6318)]

Ensures lerna does not build private packages/projects


NWRichmond (approved):
Awesome! The time to `npm run build` has gone down from minutes to seconds. Thanks for this PR, @alexcasche :)

PR #192, opened 2022-07-22
chore: scaffold changesets

alexcasche:

Fixes [ENG-6342](https://nacelle.atlassian.net/browse/ENG-6342) 

- Scaffolds changesets

### Notes
The only option that was adjusted within changesets `config.json` was to flip `"access": "public"`.  Seems like the recommended setting for public repos.  I'm not sure that we need to update any of the other default settings.  Open to feedback.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@2.1.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": false,
+  "fixed": [],
+  "linked": [],
+  "access": "public",
+  "baseBranch": "main",
+  "updateInternalDependencies": "patch",
+  "ignore": []
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,9 @@
       "name": "@nacelle/nacelle-js",
       "version": "0.0.0",
       "license": "Apache-2.0",
+      "dependencies": {
+        "@changesets/cli": "^2.24.0"
+      },
       "devDependencies": {
         "@typescript-eslint/eslint-plugin": "^5.10.1",
         "@typescript-eslint/parser": "^5.10.1",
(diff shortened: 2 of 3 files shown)


.changeset/config.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@2.1.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
```
NWRichmond, line 3: I went down a little bit of a rabbit hole learning about the `changelog` option. Here's what I found:

The [changesets docs](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#changelog-false-or-a-path) say:

> As well as the default one, you can use `@changesets/changelog-git`, which adds links to commits into changelogs, or `@changesets/changelog-github`, which requires github authentication, and includes a thankyou message to the person who added the changeset as well as a link to the relevant PR.

At first glance, that makes it sound like `@changesets/cli/changelog` doesn't add links to commits, which I know has been a pain point in some of our older changesets-powered changelogs ([example](https://github.com/getnacelle/compatibility-connector/blob/main/CHANGELOG.md#nacellecompatibility-connector)).

That said, [changes were made to `@changesets/cli` in February](https://github.com/changesets/changesets/pull/666/files) that make `@changesets/changelog-git` the default. Technically, `@changesets/cli/changelog` is now an alias for `@changesets/changelog-git` as of [`@changesets/cli@2.21.0`](https://github.com/changesets/changesets/pull/750/files#diff-6e2e2a1851648938b325ba84de634407a4e69a644ea61102df15ca4a8a7a9758).

So I think the default should be good, as it will provide [links to commits in the changelog](https://github.com/changesets/changesets/blob/main/packages/changelog-git/CHANGELOG.md#changesetschangelog-git). At this point, the only question is whether or not we see value in including a thank you message to contributors in the changelog, [which `@changesets/changelog-github` accomplishes](https://github.com/changesets/changesets/blob/main/packages/changelog-github/CHANGELOG.md#changesetschangelog-github):

changesets changlog github pr thanks to contributors

NWRichmond (approved):
Looks good! Might be worth discussing the `changelog` setting as a group before finalizing. But whatever the decision there, we should be in a good position either way.

alexcasche:
> Looks good! Might be worth discussing the `changelog` setting as a group before finalizing. But whatever the decision there, we should be in a good position either way.

Makes sense.  This was really just an initial stab to get us unblocked on the epic.  Definitely worth discussing as a group

PR #208, opened 2022-08-22
feat(ENG-6319): sanity plugin nacelle references

alexcasche:

[ENG-6319](https://nacelle.atlassian.net/browse/ENG-6319) 

Adds support for Nacelle Reference objects per [ADR040](https://nacelle.atlassian.net/wiki/spaces/DD/pages/1749712897/ADR+040+-+Structured+Commerce+Data+References+in+Content)

- Adds a second input/part to plugin export `sanity.json`
- Refactors/renames/moves reusable components to `shared` folder
- Renames top level component `nacelleData` to match package exports
- updates HF queries to include `__typename`
- Creates new top level component/schema type `nacelleReference` for storing reference objects

## How to Test
For local testing I used [Sanity Reference Store](https://github.com/getnacelle/nacelle-reference-store-sanity-studio) repo.

- Pull down this branch, build and link the sanity plugin to local Sanity studio project
- Update a content schema in Sanity studio project to use `nacelleReference` type (see below).  For this example we used `schemas/product/_content.js`
Screen Shot 2022-08-22 at 10 25 04 AM
- Update the piece of content so that it links to a new product
Screen Shot 2022-08-22 at 10 27 04 AM
- Publish content, open the vision tab, and query for it.  You should see the `NacelleReference` object stored in the response.
Screen Shot 2022-08-22 at 10 28 01 AM

Notes:  The Sanity UI for both of the inputs intentionally looks identical.  We could show _what_ is being stored, but it felt a bit unnecessary.  Right now the saving of the object occurs behind the scenes.


NWRichmond (changes requested):
Great work overall! The refactor is clean.

I think there are only two outstanding issues:
1. The formatting of `referenceType`. I left a detailed comment about this on `packages/sanity-plugin-nacelle-input/src/components/NacelleReference.js`.
2. We should provide guidance in the README. I think promoting `nacelleReference` as the go-to type would be wise, so long as we point out that `nacelleData` should be used for backwards-compatibility with content that contains references created with [`@nacelle/sanity-plugin-pim-linker`](https://www.npmjs.com/package/@nacelle/sanity-plugin-pim-linker).

Other than that, everything looks great. In local testing with a fresh Sanity project, the `nacelleData` references came back as strings and were totally compatible with existing entries that contained `nacelleData` references from `@nacelle/sanity-plugin-pim-linker`. And the `nacelleReference`s contained the expected properties.

For curious minds, this is the UI that Sanity presents when you change an existing `nacelleData` field (that contains data) to a `nacelleReference` field:

Sanity setting previous handle field to structured reference


All is well after you clear the data and replace it with a `nacelleReference` :)

alexcasche:
@NWRichmond should be all set here now.  Couple notes:
1.  Updated `README.md` to focus on the `nacelleReference` type.  We will likely want to add something around backend data joins when that is released.
2. I opted to use an Object Literal to map the `referenceType` values so it was a little more clear for future devs ```{ 'Product': 'PRODUCT', 'ProductCollection': 'COLLECTION}```

packages/sanity-plugin-nacelle-input/README.md
```diff
@@ -91,33 +91,53 @@ Realistically, you probably want to restrict the component to _either_ products
...
+  type: 'NacelleReference',
+  referenceType: 'PRODUCT', // PRODUCT or COLLECTION
+  nacelledEntryId: '123455'
+  handle: 'blue-shirt'
+  locale: 'en-US'
+}
+```
 
-Since this custom input component just stores the `handle` of a particular product or collection, you'll use the [Nacelle Storefront SDK](https://nacelle.com/docs/querying-data/storefront-sdk?nacelle=v2&storefrontSdk=1.0.3) to fetch the associated product or collection object.
+## Backwards compatibility
```
NWRichmond, line 117:
```suggestion
## Backward compatibility with `@nacelle/sanity-plugin-pim-linker`
```

packages/sanity-plugin-nacelle-input/README.md
```diff
@@ -91,33 +91,53 @@ Realistically, you probably want to restrict the component to _either_ products
...
 
-#### Product
+A backwards compatible custom input called `nacelleData` is also inlcuded. This input stores only the `handle` of a particular product or collection.
```
NWRichmond, line 119: I think we should be explicit about what this is backwards-compatible with:

```suggestion
For projects that were using [@nacelle/sanity-plugin-pim-linker](https://www.npmjs.com/package/@nacelle/sanity-plugin-pim-linker) for product/collection references, we provide a backward-compatible custom input called `nacelleData`. This input stores only the `handle` of a particular product or collection rather than a structured `nacelleReference` object.
```

NWRichmond (approved):
Awesome, thanks for those updates, @alexcasche!

I added one more suggestion re: `nacelleData` docs:

PR #205, opened 2022-08-16
feat: nuxt-3 starter example

mdarrik:
## WHY are these changes introduced?

Example of Nuxt 3 - [ENG-4185](https://nacelle.atlassian.net/browse/ENG-4185) 

## WHAT is this pull request doing?

Adds an example of Nuxt 3 + Nacelle, using the existing Nuxt starter as a template.

- Features a composable-based state management solution instead of Vuex/Pinia.
- Shopify Cart/Checkout
- Nuxi generate support with pre-rendered routes for a static experience.
- The ReadMe also links out to relevant issues around static generation support for tracking of when generate/hybrid modes might be stable.

## How to Test

🖥 [Local development instructions](https://github.com/getnacelle/nacelle-js/blob/mdarrik/ENG-4185-Nuxt-3-starter-example/examples/nuxt-3/README.md)

1. Checkout mdarrik/ENG-4185-Nuxt-3-starter-example (or using gh CLI: `gh pr checkout 205`)
2. Run `npm run bootstrap` from root of the monorepo
3. Navigate to examples/nuxt-3/
4. Copy the `.env.example` file into a `.env` file and fill out the values (I can provide the ones I used if desired)
5. Run dev server with `npm run dev` and test pages and cart/checkout
6. Do a production static build with `npm run generate`
7. Test the build with `npx serve .output/public` and navigate to pages and test cart/checkout

PR #209, opened 2022-08-23
fix: linting

NWRichmond:

To support #205.

Currently, it doesn't seem possible to run Nuxt 2 and Nuxt 3 linting setups in the same monorepo. To do that, [Volar](https://marketplace.visualstudio.com/items?itemName=Vue.volar) would need to provide an option to define sub-projects that it runs in, [as Vetur does](https://vuejs.github.io/vetur/guide/setup.html#advanced).

With that in mind, it's challenging to run ESLint on the `examples/nuxt-3` project introduced in #205. We can do this locally, though, using an ESLint config setup [that closely mirrors this](https://github.com/nuxt/framework/discussions/2815#discussioncomment-2050408).

This PR fixes some issues highlighted by:
1. `'eslint:recommended'`
2. `'plugin:vue/vue3-essential'`
3. `'plugin:@typescript-eslint/recommended'`
4. `'@nuxtjs/eslint-config-typescript'`

Most of them are superficial (formatting, unused variables, default props, etc.). The more consequential fixes are:
- `examples/nuxt-3/composables/cart/index.ts` had a variable name collision that led to a problematic comparison: `item.id === item.id`
- In `examples/nuxt-3/pages/collections/[handle].vue`, there was a dangerous optional chaining that could result in a `TypeError`, if `allProductCollections.edges[0]` was `undefined`.

### DIFF
--- a/examples/nuxt-3/pages/collections/[handle].vue
+++ b/examples/nuxt-3/pages/collections/[handle].vue
@@ -42,7 +42,7 @@ const { data: collectionPageData, error: pageFetchError } = await useAsyncData(
       const {
         productConnection: {
           pageInfo: productPaginationInfo,
-          productCount,
+          totalCount: productCount,
           edges: productEdges
         },
         ...productCollection
--- a/examples/nuxt-3/pages/collections/[handle].vue
+++ b/examples/nuxt-3/pages/collections/[handle].vue
@@ -42,17 +42,17 @@ const { data: collectionPageData, error: pageFetchError } = await useAsyncData(
       const {
         productConnection: {
           pageInfo: productPaginationInfo,
-          totalProductCount,
+          productCount,
           edges: productEdges
         },
-        ...collection
-      } = allProductCollections.edges[0]?.node;
-      const products = productEdges.map((edge) => edge.node);
+        ...productCollection
+      } = allProductCollections.edges[0].node;
+      const collectionProducts = productEdges.map((edge) => edge.node);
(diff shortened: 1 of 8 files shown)


examples/nuxt-3/pages/collections/[handle].vue
```diff
@@ -42,17 +42,17 @@ const { data: collectionPageData, error: pageFetchError } = await useAsyncData(
...
-          totalProductCount,
+          productCount,
           edges: productEdges
         },
-        ...collection
-      } = allProductCollections.edges[0]?.node;
-      const products = productEdges.map((edge) => edge.node);
+        ...productCollection
+      } = allProductCollections.edges[0].node;
+      const collectionProducts = productEdges.map((edge) => edge.node);
```
Comment (mdarrik) [inline examples/nuxt-3/pages/collections/[handle].vue:50]: It's worth calling out that this still errors  if allProductCollections.edges[0] is undefined. It just doesn't have any eslint hints since it's untyped (which might be okay).

Comment (NWRichmond) [reply examples/nuxt-3/pages/collections/[handle].vue:50]: Come to think of it, since we're in a:
```js
if (allProductCollections.edges[0]) {
```
, we should be safe.

mdarrik (approved):
LGTM

examples/nuxt-3/pages/collections/[handle].vue
Comment (mdarrik) [reply examples/nuxt-3/pages/collections/[handle].vue:50]: Good catch!

PR #210, opened 2022-08-24
chore: workflow + scripts

alexcasche:

[ENG-6820](https://nacelle.atlassian.net/browse/ENG-6820) 

- Updates release workflow, and adds postinstall bootstrap

### Testing

- If release workflow succeeds we should be good

_I don't think we need a changeset for this PR since no packages should get bumped?_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


mdarrik (approved):
Workflow changes seem like they should work. And the postinstall script worked for me! LGTM

PR #224, opened 2022-08-31
feat: update mocks & tests to not use encoded gid

KrisQ:

Fixes [[ENG-6954](https://nacelle.atlassian.net/browse/ENG-6954)] 

Updates tests to use regular global ids instead of base64 encoded global ids for cart

## 

### How to Test

- `git checkout ENG-6954--global-id`
- `cd packages/shopify-cart`
- `npm run test:ci`


mdarrik (approved):
LGTM - tests all pass. Might have a small merge conflict with #223, but I'll handle that if it comes up.

alexcasche (approved):
Nicely done!

PR #227, opened 2022-09-02
refactor(ENG-7006): update transformLineItem functions & add test

alexcasche:

Addresses [ENG-7006](https://nacelle.atlassian.net/browse/ENG-7006)

> Refactors `utils/transformNacelleLineItemToShopifyLineItem`

- Simplifies the type assertions and `if` blocks in `transformNacelleLineItemToShopifyLineItem`
- Pulls out `nacelleEntryId`, but passes along all the other user supplied properties.  This is important for two reasons.  1.  It ensures we don't need to update this function anytime the mutations in Shopify accept new params.  2.  It ensures users are given Shopify errors if they attempt to pass in properties that don't exist.
- Adds test to get to 100% coverage

## How to test
- `git checkout ENG-7006/additional-util-testing`
- `cd packages/shopify-cart`
- `npm run test:ci`

Screen Shot 2022-09-02 at 11 12 20 AM


.changeset/smooth-walls-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nacelle/nacelle-js": patch
+---
+
+feat(breaking!): update transformLineItem functions & add tests
```
mdarrik, line 5: I'm not sure if this is breaking.

mdarrik (approved):
Nice! I hated how complex it was before this. Thanks for figuring out a better way.

.changeset/smooth-walls-grin.md
alexcasche: I wasn't sure.  I thought it might be because users would now get errors they hadn't before.  Happy to update though

NWRichmond (approved):
Thanks for this!

Everything is working as expected. If you think there is value in simplifying the types (and consequently, the `actions`) further, I put opened #228, which leverages [TS conditional types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html). Whether or not that PR is merged, I think this PR is on solid footing.

.changeset/smooth-walls-grin.md
alexcasche: @NWRichmond - since the transform function now passes along _all_ the parameters supplied by users, it means if incorrect parameters are supplied, Shopify errors will be returned.  Before this update, we pruned only the accepted params and passed them along.  I was on the fence here between breaking and not breaking

NWRichmond: Ah okay, that makes sense. In that case, updating this changeset to describe that scenario would be helpful.

.changeset/smooth-walls-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nacelle/nacelle-js": patch
+---
+
+feat(breaking!): cartLines now require correct shopify params.  Incorrect params will now throw an error
```
mdarrik: Was there a reason incorrect params wouldn't throw before? They were still getting passed to the actual fetch call

PR #228, opened 2022-09-02
refactor: conditional types to simplify actions

NWRichmond:

To support #227.

Refactors `packages/shopify-cart/src/utils/transformNacelleLineItemToShopifyLineItem.ts` with [conditional types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html). This allows us to eliminate type assertions where `transformNacelleLineItemToShopifyLineItem` is called, which in turn allows us to simplify some of the `packages/shopify-cart/src/client/actions`.

The idea with the conditional types is that by checking to see if the input to `transformNacelleLineItemToShopifyLineItem` is `NacelleCartLineItemUpdateInput` or not (when it's not, it's `NacelleCartInput`), we can have TypeScript automatically know if the `transformNacelleLineItemToShopifyLineItem` return value is `CartLineInput` or `CartLineUpdateInput`.

For example (keeping in mind that the distinction between `NacelleCartInput` and `NacelleCartLineItemUpdateInput` is that `NacelleCartLineItemUpdateInput` has an `id: string` property)...

`transformNacelleLineItemToShopifyLineItem` correctly returns `CartLineInput[]` when its input doesn't have an `id: string`:

transformNacelleLineItemToShopifyLineItem infer CartLineInput type

, and when `transformNacelleLineItemToShopifyLineItem` does have an `id: string`, it correctly returns `CartLineUpdateInput[]`:

transformNacelleLineItemToShopifyLineItem infer CartLineUpdateInput type

## 

## How to Ttest

1. Check out the `ENG-7006/conditional-types` branch.
2. Navigate to `packages/shopify-cart`.
3. `npm run test:ci` - all tests should pass with 100% coverage.

ENG-7006:conditional-types all tests passing

5. `npm run build` - the package should build without errors.

ENG-7006:conditional-types builds without errors

6. `npm run codegen` - there should be no new changes to `packages/shopify-cart/src/types/shopify.type.ts`.

### DIFF
--- a/packages/shopify-cart/src/client/actions/cartCreate.ts
+++ b/packages/shopify-cart/src/client/actions/cartCreate.ts
@@ -33,20 +33,25 @@ export default async function cartCreate({
   gqlClient,
   params
 }: CreateCartParams): Promise<void | CartResponse> {
+  let shopifyParams: CartInput = {};
+
   try {
-    let formattedParams: NacelleCartInput | CartInput | undefined = params;
-    if (params && params.lines) {
-      const nacelleLines = params.lines;
-      const shopifyLines =
-        transformNacelleLineItemToShopifyLineItem(nacelleLines);
-      formattedParams = { ...params, lines: shopifyLines } as CartInput;
+    if (params) {
+      const { lines: nacelleLines, ...otherParams } = params;
+      shopifyParams = otherParams;
+
+      if (nacelleLines) {
+        shopifyParams.lines =
+          transformNacelleLineItemToShopifyLineItem(nacelleLines);
+      }
     }
+
     const shopifyResponse = await gqlClient<
       MutationCartCreateArgs,
       MutationCartCreateResponse
     >({
       query: mutations.CART_CREATE(customFragments),
-      variables: { input: (formattedParams as CartInput | undefined) ?? {} }
+      variables: { input: shopifyParams }
     }).catch((err) => {
       throw new Error(err);
     });
--- a/packages/shopify-cart/src/utils/transformNacelleLineItemToShopifyLineItem.ts
+++ b/packages/shopify-cart/src/utils/transformNacelleLineItemToShopifyLineItem.ts
@@ -1,13 +1,17 @@
-import {
-  NacelleCartLineItemInput,
-  NacelleCartLineItemUpdateInput
-} from '../types/cart.type';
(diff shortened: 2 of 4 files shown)


packages/shopify-cart/src/client/actions/cartCreate.ts
```diff
@@ -33,20 +33,25 @@ export default async function cartCreate({
   gqlClient,
   params
 }: CreateCartParams): Promise<void | CartResponse> {
+  let shopifyParams: CartInput = {};
```
NWRichmond, line 36: @mdarrik I'd love your take on these changes, given our discussion in https://github.com/getnacelle/nacelle-js/pull/223/files/e755b61c8eb358dfbccf697bde6e07b385d3c843#r960042889 about subtle ways that this action can go sideways.

packages/shopify-cart/src/utils/transformNacelleLineItemToShopifyLineItem.ts
```diff
@@ -1,19 +1,23 @@
...
+import type { NacelleCartLineItemUpdateInput } from '../types/cart.type';
+import type { CartLineInput, CartLineUpdateInput } from '../types/shopify.type';
 
-export default function (
-  nacelleLines: NacelleCartLineItemInput[] | NacelleCartLineItemUpdateInput[]
-): CartLineInput[] | CartLineUpdateInput[] {
+/**
+ * If type input `T` is an array of objects containing properties of `NacelleCartLineItemUpdateInput`, we'll return `CartLineUpdateInput[]`. Otherwise, we'll return `CartLineInput[]`.
+ */
+type InputOrUpdateInput<T> = T extends NacelleCartLineItemUpdateInput[]
```
alexcasche, line 8: This is super nifty

alexcasche (approved):
Really neat trick

packages/shopify-cart/src/client/actions/cartCreate.ts
mdarrik: This does resolve the issue I pointed out there. Namely you're assigning all of the params to `shopifyParams` whether or not the `params.lines` exist. It's got that same issue of doing extra work to make TS happier, but it is a lot less type assertions.

NWRichmond: Thank you for the insights!

PR #234, opened 2022-09-08
docs: add `customFragments` to client init options

NWRichmond:

Addresses [ENG-7063](https://nacelle.atlassian.net/browse/ENG-7063)

> Add `customFragments` to "Initializing the client" docs in the README

In the "Initializing the client" docs, `customFragments` were missing from the list of initialization options.

- Adds `customFragments` to the list of initialization options.
- Corrects the type of `fetchClient` (was `object`, should be `function`).

Here's what the README looks like now:

@nacelle:shopify-cart initialization options with customFragments


KrisQ (approved):
Looks good!

PR #238, opened 2022-09-12
feat(shopify-cart): Adds params for setting country and language codes

mdarrik:

Addresses [ENG-7060](https://nacelle.atlassian.net/browse/ENG-7060) - Adds parameters to `cartCreate` for setting the `language` and `country` in all Cart queries and mutations. This allows the cart to support internationalized stores. 

1. Adds `language` and `country` params to the `createCart` function that correspond to the Shopify GQL [Language Code][shopify-cart-language-code] and [Country Code][shopify-cart-country-code] enums. If not provided, the language code defaults to `EN` and the country code corresponds to `ZZ`. Note that prior to this PR, no language code was used and all queries used `ZZ` for the country code.
2. Updates All Shopify Enums to be types - this makes them easily accessible as strings, which makes a better public API for users. 
3. Updates every query/mutation GQL to use accept a `language` variable parameter, and to use the `language` in the `inContext` directive.
4. Passes the `language` & `country` variables to every action, and uses them in the queries. Updates the types passed to the `GQLClient` in each action to use the `<QueryName>Variables` type so that `language` and `country`  are expected. This should allow us to update query variables without needing to change types in the future too. 
5. Adds `language` and `country` to the args/responses for all tests.
6. Adds a test for passing a non-default language to `createCart` & a test passing a non-default `country` to `createCart`
7. Updates the `README` to document the 2 new parameters to `createCart`. 

## How to Test

1. Clone this branch `mdarrik/shopify-cart/ENG-7060-add-params-for-setting-and-language` (or `gh pr checkout #` if using the gh cli)
2. Run `npm run test:ci` - all tests should passed
3. Run `npm run build` - build should complete successfully
4. Use `npm link` to test a local project featuring shopify cart. Outgoing queries should include the `language` and `country` specified, or default to `EN` and `ZZ` respectively.

[shopify-cart-language-code]: https://shopify.dev/api/storefront/2022-07/enums/LanguageCode
[shopify-cart-country-code]: https://shopify.dev/api/storefront/2022-07/enums/CountryCode

### DIFF
--- a/.changeset/cold-oranges-study.md
+++ b/.changeset/cold-oranges-study.md
@@ -1,0 +1,7 @@
+---
+'@nacelle/shopify-cart': minor
+---
+
+feat (breaking): Adds options for language and country codes to cart create.
+
+- Now defaults to setting the language to `EN` in all queries and mutations if no language is passed to the `createCart` function.
--- a/packages/shopify-cart/README.md
+++ b/packages/shopify-cart/README.md
@@ -61,6 +61,8 @@ The cart client accepts the following parameters:
 - `shopifyCustomEndpoint` (`string`) - optional Shopify custom API endpoint
 - `customFragments` (`object`) - optional GraphQL fragments for data customization (see the [Custom Fragments docs][custom-fragments-docs])
 - `fetchClient` (`function`) - optional Fetch API-compatible request function
+- `country` (`string`) - optional [Shopify Country Code][shopify-cart-country-code] used for cart queries/responses. Defaults to `ZZ` (unknown region) if not supplied.
+- `language` (`string`) - optional [Shopify Language Code][shopify-cart-language-code] used for cart queries/responses. Defaults to `EN` if not supplied.
 
 _You must provide a `shopifyShopId` or a `shopifyCustomEndpoint`_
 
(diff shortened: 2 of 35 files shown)


KrisQ (approved):
Looks good & works!

Screen Shot 2022-09-13 at 7 29 53 AM

Screen Shot 2022-09-13 at 7 29 31 AM

alexcasche (commented):
I think we just need to align around how we use the `enums` as Nick called out.  Since most of our merchants are in `US`, would that be a better default for country code over `ZZ`?  I think `unknown` makes more sense, but setting `US` _could_ save majority of merchants from passing that param.

mdarrik:
I've updated the enums generated by codegen to be types. This makes them nice, user-friendly strings and fixes the TS issues when trying to use this branch.

mdarrik:
@alexcasche you might know better, but it seems like country code of ZZ doesn't affect merchants who have a US store only? We were passing ZZ in all of our queries before this branch.

alexcasche:
> @alexcasche you might know better, but it seems like country code of ZZ doesn't affect merchants who have a US store only? We were passing ZZ in all of our queries before this branch.

Sounds right to me!  Thanks for clarifying

NWRichmond (approved):
Excellent! Thanks for this PR, @mdarrik.

`build`, `codegen`, and `test:ci` are all working as expected ✅

The conversion of enums to string literals is a very nice touch!

@nacelle:shopify-cart type import of language as string literal

PR #245, opened 2022-09-14
chore: Configure Changesets to publish Shopify Cart Beta

mdarrik:
**NOTE:** This PR is a draft PR until we're ready to release the 1.0.0-beta. 

Addresses [ENG-7143](https://nacelle.atlassian.net/browse/ENG-7143) - sets up changesets and the gh actions for releasing the Shopify Cart 1.0.0-beta 

1. Adds missing "version" key to `examples/nuxt-3/package.json`. This fixes an error preventing `changeset version` from working in pre-release mode
2. Updates the "name" key in `reference-store/next/package.json` to be `next-reference-store` instead of `next` since this confused changesets into thinking all NextJS repos were using the reference store as a dependency.
3. Updates any changesets for breaking changes to be tagged as `major` instead of `minor` so that changeset knows to up the version to `1.0.0-beta` instead of `0.X.X-beta`
4. Enters changeset's pre-release mode with the `beta` tag. This will cause `changesets version` to publish the version as `1.0.0-beta`
5. Adds `ENG-6843/shopify-cart-1.0.0` as a branch to trigger the `release` workflow from. (Uncertainty: I'm not 100% certain if updating it here will work or if I'll need to update it on the `main` branch via another PR. Some things work at the branch level in actions, and some have to be set at the main branch level). 

##  How to Test

There's not much testing since this is more of  a config PR, but you can test the version bump by running `npm run changeset version` from the root of the repo after checking out `mdarrik/shopify-cart/ENG-7143-configure-beta-publish-step` in git.

### DIFF
--- a/.changeset/cold-oranges-study.md
+++ b/.changeset/cold-oranges-study.md
@@ -1,5 +1,5 @@
 ---
-'@nacelle/shopify-cart': minor
+'@nacelle/shopify-cart': major
 ---
 
 feat (breaking): Adds options for language and country codes to cart create.
--- a/.changeset/gorgeous-gifts-sip.md
+++ b/.changeset/gorgeous-gifts-sip.md
@@ -1,5 +1,5 @@
 ---
-"@nacelle/nacelle-js": patch
+"@nacelle/shopify-cart": patch
 ---
 
 docs: support/release schedule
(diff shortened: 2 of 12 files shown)


KrisQ (approved):
Looks good!

PR #248, opened 2022-09-16
feat: adds get/set config functions

mdarrik:

Addresses [ENG-7167](https://nacelle.atlassian.net/browse/ENG-7167) - Adds a setConfig and getConfig to the cart client. 

- Adds a getConfig method that returns the string options passed to `createCartClient`
- Adds a `setConfig` method to allow you to update the `language` and `countryCode`

## How to Test

- Checkout this branch `mdarrik/shopify-cart/ENG-7167-add-setConfig-properties-to-cart` 
- Run `npm run test:ci` - all tests should pass
- Run `npm run build` - should build successfully
- To test in a local repo - use `npm link` and try calling `getConfig` to see the options used in the cart, and test updating the language/country with `setConfig`

### DIFF
--- a/.changeset/old-colts-attend.md
+++ b/.changeset/old-colts-attend.md
@@ -4,5 +4,5 @@
 
 feat: adds `getConfig` and `setConfig` functions
 
-- `getConfig` allows for reviewing options set in the cart client such as the `language`, `country`, the shopify Shop Id, and other options.
+- `getConfig` allows for reviewing options set in the cart client such as the `language`, `country`, the Shopify Shop Id, and other options.
 - `setConfig` allows for updating some of the configuration options used to make requests to Shopify - namely `language` and `country` at this juncture.
--- a/.changeset/old-colts-attend.md
+++ b/.changeset/old-colts-attend.md
@@ -1,0 +1,8 @@
+---
+'@nacelle/shopify-cart': minor
+---
+
+feat: adds `getConfig` and `setConfig` functions
+
+- `getConfig` allows for reviewing options set in the cart client such as the `language`, `country`, the shopify Shop Id, and other options.
+- `setConfig` allows for updating some of the configuration options used to make requests to Shopify - namely `language` and `country` at this juncture.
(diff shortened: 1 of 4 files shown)


NWRichmond (commented):
The `build` and `test:ci` scripts are working as expected ✅

Running the `codegen` script results in uncommitted changes, but they all appear to be changes on Shopify's end (adding periods to the end of comments), and not related to changes made in this PR. Feel free to run `codegen` again and commit changes, or we can handle that on the [`ENG-6843/shopify-cart-1.0.0`](https://github.com/getnacelle/nacelle-js/tree/ENG-6843/shopify-cart-1.0.0) branch.

The `setConfig` is setting the `country` and `language` config as expected:

https://user-images.githubusercontent.com/5732000/191074102-ea3d0521-3985-456f-b161-4e19ed293389.mov

Personally, I find the results of `getConfig` somewhat unintuitive. I think it might make more sense to only return properties from `getConfig` that are settable in `setConfig`. I think it'd be perfectly fine to remove `shopifyShopId` from `shopifyStorefrontAccessToken` from the `getConfig` results, as this is something that we probably shouldn't expect devs to need to check in their code.

mdarrik:
> Personally, I find the results of `getConfig` somewhat unintuitive. I think it might make more sense to only return properties from `getConfig` that are settable in `setConfig`. I think it'd be perfectly fine to remove `shopifyShopId` from `shopifyStorefrontAccessToken` from the `getConfig` results, as this is something that we probably shouldn't expect devs to need to check in their code.

Yeah. I was sort of conflicted on this part of the `getConfig` piece. My thought on keeping it was that it provides a way to provide debugging/setup information which could be useful for merchant developers trying to debug setup, or for merchants with multiple shopify storefronts to cover their different locals. But happy to drop those fields if we think that's not all that useful/beneficial.

NWRichmond:
Ah, good to know. I think removing them from the `getConfig` response is probably the right move for now. We can always reevaluate that in response to feature requests etc.

> or for merchants with multiple shopify storefronts to cover their different locales

For what it's worth, because of the nature of [Shopify Markets](https://help.shopify.com/en/manual/markets), merchants should be able to accomplish international support via a single Shopify store (& token).

mdarrik:
@NWRichmond removed those fields from `getConfig`. Also updated the docs to include both methods. 

While I was in there I added an href to all of the method `a` tags so they are [valid links](https://twitter.com/BenDMyers/status/1571849889155260417). If we don't want them to be, we should switch them to spans instead.

NWRichmond (approved):
Awesome! Thanks for those updates, @mdarrik. Everything is working as expected.

shopify cart getConfig returns language and country

PR #250, opened 2022-09-20
refactor(ENG-7181): transform nacelleEntryId updates

alexcasche:

Addresses [ENG-7181](https://nacelle.atlassian.net/browse/ENG-7181)

> Generate `nacelleEntryIds` from additional `createShopifyCartClient` params `locale, shopifyShopId`

Because of some issues with storing `nacelleEntryId` on as a line item attribute, we need a new solution for returning the cart with `nacelleEntryIds`.  To ensure we are able to transform shopify gid to a nacelleEntryId, we need access to two values `shopifyShopId` and `locale`.  This PR touches a lot of files so take your time reviewing.

### How to Test

`git checkout ENG-7181/shopify-cart-nacelle-entry-ids`
`npm run test`
Screen Shot 2022-09-20 at 12 17 25 PM

#### e2e testing

I tested this branch using a demo vite project.  All methods seem to work well.  Feel free to e2e test against any project you see fit.


packages/shopify-cart/src/utils/transformShopifyLineToNacelleLine.ts
```diff
@@ -1,15 +1,28 @@
...
     return {
-      ...transformedLine,
+      ...line,
       merchandise: {
-        ...transformedLine.merchandise,
-        nacelleEntryId
+        ...line.merchandise,
+        nacelleEntryId: btoa(
+          `id://SHOPIFY/${shopifyShopId}/default/PRODUCT_VARIANT/${shopifyVariantId}/${locale}`
+        )
```
mdarrik, line 25: Did we want to handle Node differently? I think we determined that the btoa/atob methods should be sufficient for our use case, but wanted to verify.

NWRichmond: I wish we could, but unfortunately we can't rely on Node's `atob`, for a couple of reasons:
1. It was introduced in Node `16.0.0`, and we have customers that build their projects with Node 14.
2. For reasons that aren't clear to me, `atob` and `btoa` are listed as [Stability 3: Legacy](https://nodejs.org/docs/latest-v18.x/api/all.html#all_documentation_stability-index). Both Node 16 and Node 18 docs suggest that people use `Buffer.from(data, 'base64')` instead.

Node 16 atob stability 3 legacy

Node 18 atob btoa legacy

NWRichmond: So @alexcasche we should implement an isomorphic base64 decoding. For example: https://github.com/getnacelle/nacelle-js/blob/ENG-6843/shopify-cart-1.0.0/packages/shopify-cart/src/utils/getShopifyIdFromNacelleId.ts#L8-L12

Since we'd be doing that in multiple places, it might make sense to have a single utility function that we can use wherever decoding needs to happen.

mdarrik: Worth calling out that that line is decoding a base64 encoded gid and the one here encoding one. This might be the only place we're base64 encoding anything, since all the shopify stuff is decoded by default.

mdarrik (commented):
So all of the code seems good. But other than the atob/btoa stuff that's addressed, it'd probably be good to have a test (or a few) that test other locales besides `en-US`. Just to make sure the logic is working appropriately.

NWRichmond (changes requested):
Overall, this is excellent! In local testing, the `nacelleEntryId`s returned in Shopify responses match the expected values (validated by inspecting product data in the Nacelle Dashboard).

There are just a couple of minor areas that could use attention:
1. The [current base64 encoding](https://github.com/getnacelle/nacelle-js/pull/250/files#r980473728) in `utils/transformShopifyLineToNacelleLine.ts` isn't safe for Node 14, which some of our customers still use.
2. In the **Initializing the client** section of the README, we should add `locale` to the list of params with a brief explanation that the `locale` is just used to assign `NacelleCartMerchandise` the correct `nacelleEntryId`.

packages/shopify-cart/src/utils/transformShopifyLineToNacelleLine.ts
NWRichmond: Totally. I edited the comment above to remove the suggestion to pull it out into a utility function.

alexcasche:
@NWRichmond @mdarrik I think this should be all set now.  Thanks for the reviews!

packages/shopify-cart/README.md
```diff
@@ -61,6 +61,7 @@ The cart client accepts the following parameters:
 - `fetchClient` (`function`) - optional Fetch API-compatible request function
 - `country` (`string`) - optional [Shopify Country Code][shopify-cart-country-code] used for cart queries/responses. Defaults to `ZZ` (unknown region) if not supplied.
 - `language` (`string`) - optional [Shopify Language Code][shopify-cart-language-code] used for cart queries/responses. Defaults to `EN` if not supplied.
+- `locale` (`string`) - optional [Nacelle Locale Code]used to assign `NacelleCartMerchandise` the correct `nacelleEntryId`. Defaults to `en-US` if not supplied.
```
NWRichmond, line 64: @alexcasche this isn't quite right yet

Screen Shot 2022-09-27 at 11 16 03 AM

packages/shopify-cart/src/utils/transformShopifyLineToNacelleLine.ts
```diff
@@ -1,14 +1,33 @@
...
+    const shopifyVariantId = line.merchandise.sourceEntryId
+      .split('gid://shopify/ProductVariant/')[1]
+      .replace('/', '');
+
+    let nacelleEntryId = `id://SHOPIFY/${shopifyShopId}/default/PRODUCT_VARIANT/${shopifyVariantId}/${locale}`;
+
+    if (typeof window !== 'undefined') {
+      nacelleEntryId = globalThis.btoa(nacelleEntryId);
+    } else {
+      nacelleEntryId = new Buffer(nacelleEntryId).toString('base64');
```
NWRichmond, line 24:
```suggestion
      nacelleEntryId = Buffer.from(nacelleEntryId).toString('base64');
```

buffer is deprecated

From https://nodejs.org/en/docs/guides/buffer-constructor-deprecation:

new Buffer security risk

NWRichmond (changes requested):
Sorry, one last change request!

NWRichmond (approved):
Awesome! Thanks for sticking with this one, @alexcasche. I know it was tricky 😅

PR #252, opened 2022-09-20
chore: rename prepare to nuxt-prepare so ci succeeds

mdarrik:

Fixes CI by renaming the `prepare` script in the `examples/nuxt-3` `package.json`. `prepare` is an npm lifecycle script which we don't want to trigger during regular `ci` installs. 

## How to Test

1. Checkout this branch `mdarrik/shopify-cart/rename-prepare-script
2. Remove/rename any local `.env` file in the `examples/nuxt-3` folder && remove the `.nuxt` folder.
3. Run `CI=true npm ci` from the repo root - should succeed.


NWRichmond (approved):
Thanks!

PR #255, opened 2022-09-21
fix: validate cart responses exist

alexcasche:

Addresses [ENG-7136](https://nacelle.atlassian.net/browse/ENG-7136)

> Check existense of parent key before accessing child.  This was causing js errors to be returned instead of Shopify cart/error responses.

## 

### How to Test

- `git checkout ENG-7136/validate-shopify-errors`
- `npm run test:ci`

### DIFF
--- a/.changeset/strange-lemons-pay.md
+++ b/.changeset/strange-lemons-pay.md
@@ -1,0 +1,5 @@
+---
+'@nacelle/shopify-cart': minor
+---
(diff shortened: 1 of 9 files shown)


NWRichmond (approved):
Great!

The changes make sense, and `build`, `codegen`, and `test:ci` scripts are all working as expected.

PR #249, opened 2022-09-20
feat: shopify-cart in nuxt starter

KrisQ:
### WHY are these changes introduced?

Fixes [[ENG-7070](https://nacelle.atlassian.net/browse/ENG-7070)] 

### WHAT is this pull request doing?

Adds shopify cart to the nuxt starter
Replace checkout with shopify cart checkout

### How to Test

- `git checkout ENG-7070--nuxt-starter-shopify-cart`
- to the nuxt reference store project `.env` add the following: (reach out if you want the vars I used)

```
SHOPIFY_SHOP_ID="<YOUR_SHOPIFY_ID>"
SHOPIFY_STOREFRONT_ACCESS_TOKEN="<YOUR_SHOPIFY_STOREFRONT_ACCESS_TOKEN>"
```

- `cd starter/nuxt`
- `npm i && npm run generate && npm run start`
- see if cart/checkout works

PR #257, opened 2022-09-26
feat(ENG-7069)!: next starter shopify-cart

alexcasche:
Draft for now


starters/next/components/Layout.js
```diff
@@ -1,22 +1,4 @@
...
 function Layout({ children }) {
-  const [, { clearCart }] = useCart();
-  const [{ completed }, { clearCheckoutData }] = useCheckout();
-
-  useEffect(() => {
-    if (completed) {
-      clearCheckoutData();
-      clearCart();
-    }
-  }, [completed, clearCheckoutData, clearCart]);
```
NWRichmond, line 18: I can't remember what we decided re: knowing when the cart should be cleared when using Shopify Cart 🤔

NWRichmond (changes requested):
This is a good start! Here are some recommended next steps:

1. Can you `git rebase -i ENG-6843/shopify-cart-1.0.0` (or merge) so that we can use `@nacelle/shopify-cart@1.0.0-beta.x`? That should allow us to generate deploy previews.
2. Let's update `.env.example` to:

```
# Nacelle
NEXT_PUBLIC_NACELLE_STOREFRONT_TOKEN="<YOUR_NACELLE_STOREFRONT_TOKEN>"
NEXT_PUBLIC_NACELLE_STOREFRONT_ENDPOINT="<YOUR_NACELLE_STOREFRONT_ENDPOINT>"

# Shopify Cart (for details, see: https://www.npmjs.com/package/@nacelle/shopify-cart)
NEXT_PUBLIC_SHOPIFY_SHOP_ID="<YOUR_MYSHOPIFY_DOMAIN>"
NEXT_PUBLIC_SHOPIFY_STOREFRONT_ACCESS_TOKEN="<YOUR_STOREFRONT_CHECKOUT_TOKEN>"

```

3. It looks like our `<Image />` `width` and `height` props aren't quite right (distorted product images in PDPs and collection PLPs). Not specific to this PR, but would be nice to fix.
4. Cart line items don't appear to be using [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)-formatted prices.

https://user-images.githubusercontent.com/5732000/192821668-76eae40c-3ab9-409f-b8ed-e7296cd2cf8f.mov

5. There's some jank with the subtotal when incrementing / decrementing:

https://user-images.githubusercontent.com/5732000/192821967-44103cf6-cd2a-449e-a7ce-72de26057175.mov

6. Can you add a link to the cart page from PDPs? [We did this for the Nuxt Starter](https://github.com/getnacelle/nacelle-js/pull/249#pullrequestreview-1117146614), and I think it was an upgrade that could be enjoyed by all `starters/*` projects.

starters/next/components/Layout.js
alexcasche: @NWRichmond  not clear on this one either to be honest

alexcasche:
> This is a good start! Here are some recommended next steps:
> 
> 1. Can you `git rebase -i ENG-6843/shopify-cart-1.0.0` (or merge) so that we can use `@nacelle/shopify-cart@1.0.0-beta.x`? That should allow us to generate deploy previews.
> 2. Let's update `.env.example` to:
> 
> ```
> # Nacelle
> NEXT_PUBLIC_NACELLE_STOREFRONT_TOKEN="<YOUR_NACELLE_STOREFRONT_TOKEN>"
> NEXT_PUBLIC_NACELLE_STOREFRONT_ENDPOINT="<YOUR_NACELLE_STOREFRONT_ENDPOINT>"
> 
> # Shopify Cart (for details, see: https://www.npmjs.com/package/@nacelle/shopify-cart)
> NEXT_PUBLIC_SHOPIFY_SHOP_ID="<YOUR_MYSHOPIFY_DOMAIN>"
> NEXT_PUBLIC_SHOPIFY_STOREFRONT_ACCESS_TOKEN="<YOUR_STOREFRONT_CHECKOUT_TOKEN>"
> ```
> 
> 3. It looks like our `<Image />` `width` and `height` props aren't quite right (distorted product images in PDPs and collection PLPs). Not specific to this PR, but would be nice to fix.
> 4. Cart line items don't appear to be using [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)-formatted prices.
> 
>  cart.line.item.price.format.mov 
> 5. There's some jank with the subtotal when incrementing / decrementing:
> 
>  increment.decrement.subtotal.mov 
> 6. Can you add a link to the cart page from PDPs? [We did this for the Nuxt Starter](https://github.com/getnacelle/nacelle-js/pull/249#pullrequestreview-1117146614), and I think it was an upgrade that could be enjoyed by all `starters/*` projects.
> 7. Can you update the `package-lock.json` with `npm i --package-lock-only`? Currently, it's missing the `@nacelle/shopify-cart` dependency.

@NWRichmond  Want to flag that some of these issues are present in Nuxt starter.  So we may want to sweep that and create a ticket to address it.  For number 5).  This is tricky.  It has to do with Optimistic UI running updates, but then reacting to shopify responses as they come resolve.  Similar issue in nuxt.  We may need to re-evaluate how we allow for back to back actions to occur

alexcasche:
@NWRichmond @KrisQ this should be good to re-review

KrisQ:
Hey @alexcasche should we change the PDP quantity selector to be limited to 10 as well for consistency?

Screen Shot 2022-09-28 at 2 07 58 PM

alexcasche:
> Hey @alexcasche should we change the PDP quantity selector to be limited to 10 as well for consistency?
> 
> Screen Shot 2022-09-28 at 2 07 58 PM

Good call.  Just updated that

KrisQ:
> > Hey @alexcasche should we change the PDP quantity selector to be limited to 10 as well for consistency?
> > Screen Shot 2022-09-28 at 2 07 58 PM
> 
> Good call. Just updated that

@alexcasche We could still add more than 10 qty, would it make sense to make the cart quantity a number input instead of select?

Screen Shot 2022-09-28 at 2 20 44 PM

Screen Shot 2022-09-28 at 2 20 50 PM

alexcasche:
> > > Hey @alexcasche should we change the PDP quantity selector to be limited to 10 as well for consistency?
> > > Screen Shot 2022-09-28 at 2 07 58 PM
> > 
> > 
> > Good call. Just updated that
> 
> @alexcasche We could still add more than 10 qty, would it make sense to make the cart quantity a number input instead of select?
> 
> Screen Shot 2022-09-28 at 2 20 44 PM
> 
> Screen Shot 2022-09-28 at 2 20 50 PM

This would lead to same issue with Optimistic UI, where users can click the number really quickly.  What I did here was to add an 1 additional `option` to the `select` if the qty in cart was over `10`.  If not we will need to come up with a much more complex implementation.  Or ensure users can't ever add more items if `10` are in cart.  Could also be a weird experience.  We have 2 routes to pursuit:

1. We say this is good enough
2. We implement a method for tracking promises sent to Shopify's cart and only reconcile the carts when we are on our final promise in array

KrisQ:
> > > > Hey @alexcasche should we change the PDP quantity selector to be limited to 10 as well for consistency?
> > > > Screen Shot 2022-09-28 at 2 07 58 PM
> > > 
> > > 
> > > Good call. Just updated that
> > 
> > 
> > @alexcasche We could still add more than 10 qty, would it make sense to make the cart quantity a number input instead of select?
> > Screen Shot 2022-09-28 at 2 20 44 PM
> > Screen Shot 2022-09-28 at 2 20 50 PM
> 
> This would lead to same issue with Optimistic UI, where users can click the number really quickly. What I did here was to add an 1 additional `option` to the `select` if the qty in cart was over `10`. If not we will need to come up with a much more complex implementation. Or ensure users can't ever add more items if `10` are in cart. Could also be a weird experience. We have 2 routes to pursuit:
> 
> 1. We say this is good enough
> 2. We implement a method for tracking promises sent to Shopify's cart and only reconcile the carts when we are on our final promise in array

I'm ok with saying this is good enough! It doesn't really break the selection either

starters/next/components/Layout.js
NWRichmond: We figured this out elsewhere, but are sharing here for posterity's sake:

After completing a checkout with a cart's `checkoutUrl`, that `cart` is returned as `null` from Shopify.

So when the page is refreshed or the site is loading for the first time, and a cart ID is persisted in browser storage, if the value of
```js
cartClient.cart({ cartId: 'the-persisted-cart-id' });
```
is `null`, we can clear the cart.

examples/nuxt-2-composition-api/package.json.lerna_backup
```diff
@@ -0,0 +1,19 @@
...
+	},
+	"devDependencies": {
+		"@nuxt/image": "^0.6.0",
+		"@nuxtjs/pwa": "^3.3.5",
+		"fibers": "^5.0.0",
+		"sass": "^1.53.0",
+		"sass-loader": "^10.2.0",
+		"serve": "^12.0.1"
+	}
+}
```
NWRichmond, line 19: This file shouldn't be committed to source control.

NWRichmond (changes requested):
This is super close!

A few last things:

1. Let's `npm i @nacelle/shopify-cart@beta` in `starters/next`, then `npm run bootstrap` in the monorepo root to ensure we're using the correct version of `@nacelle/shopify-cart`.
2. `examples/nuxt-2-composition-api/package.json.lerna_backup` shouldn't be committed to source control.
3. There's something amiss with `availableForSale: false` variants. Let's explore that:

Variants which are `availableForSale: false` can still be added to the cart:

https://user-images.githubusercontent.com/5732000/192910871-b37159c5-3255-4bcc-b18e-ff0a5ffa07a5.mov

When we look at the response from Shopify, we can see that this unavailable item exists in the cart, with a quantity of `0`. This doesn't match the cart UI, which shows a quantity of `1`:

cart quantity 0 displayed as quantity 1

I'm open to other ideas, but one solution would be to:

- [ ] Add `disabled={!selectedVariant.availableForSale}` to the ATC button in `starters/next/pages/products/[handle].js`, to match the behavior of the ATC button in `starters/next/components/ProductCard.js`.
- [ ] In `starters/next/pages/cart.js`, `.filter((item) => item.quantity)` the `cartItems` before they are mapped in the `<ul></ul>`.

alexcasche:
@NWRichmond sorry for back and forth.  I believe this is good now.  Added in some logic for clearing the cart.

PR #258, opened 2022-09-27
chore: configure tsconfig for production

NWRichmond:

Addresses [ENG-7255](https://nacelle.atlassian.net/browse/ENG-7255)

> Configure @nacelle/shopify-cart typescript for production

Previously, unit tests had type errors because `src/**/*.spec.ts` and `__tests__` were excluded from TypeScript compilation:

unit test type errors

This PR doesn't exclude `src/**/*.spec.ts` and `__tests__` in development, and only excludes them in production (`npm run build`). We accomplish this by creating a separate `tsconfig` and using it in `npm run build`.

## How to test

1. Check out the `ENG-7255-configure-ts-for-production` branch.
2. Navigate to `packages/shopify-cart`.
3. `npm run build` - there should be no build errors.
4. Open a unit test file. There should be no type errors:

unit test no type errors

## Testing checklist

* [x] The only differences between the development & production `tsconfig`s are intended differences:

tsconfig dev and prod differences include exclude

### DIFF
--- a/packages/shopify-cart/package.json
+++ b/packages/shopify-cart/package.json
@@ -16,7 +16,7 @@
     "test": "jest --watch",
     "test:ci": "jest --silent --runInBand --collectCoverage",
     "changeset": "changeset",
-    "build": "vite build && tsc && mv dist/src dist/types",
+    "build": "vite build && tsc --project tsconfig.production.json && mv dist/src dist/types",
     "release": "npm run build && changeset publish"
   },
   "homepage": "https://github.com/getnacelle/nacelle-js/tree/main/packages/shopify-cart#readme",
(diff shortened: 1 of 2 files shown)


packages/shopify-cart/package.json
```diff
@@ -16,7 +16,7 @@
     "test": "jest --watch",
     "test:ci": "jest --silent --runInBand --collectCoverage",
     "changeset": "changeset",
-    "build": "vite build && tsc && mv dist/src dist/types",
+    "build": "vite build && tsc --project tsconfig.production.json && mv dist/src dist/types",
     "release": "npm run build && changeset publish"
```
mdarrik, line 20: Can we add a tsc command that uses the dev config? Just as a way to easily type check tests and stuff.

NWRichmond: Certainly! Added in https://github.com/getnacelle/nacelle-js/pull/258/commits/a9e27456c85109c66b64d8d13b9ec04a0a58f7aa.

mdarrik (approved):
TSC works, build works,  and tests pass. 

The dev `tsc` catches type errors in tests, and the `production` one ignores them. 


LGTM

PR #270, opened 2022-10-12
chore: move doc for sanity plugin

KrisQ:

Fixes [[ENG-7195)](https://nacelle.atlassian.net/browse/ENG-7195)] 

Move doc for sanity plugin

### Testing

- Checkout changed file
- Checkout [https://nacelle.com/docs/querying-data/nacelle-sanity-plugin-nacelle-input](https://nacelle.com/docs/querying-data/nacelle-sanity-plugin-nacelle-input)


mdarrik (approved):
Looks okay! 

Do we want to specify that it's only Nacelle v2+? And link to the v1 version? I know we didn't link to it here before, so it's totally fine if we think that we don't need/want to do that. 


You'll also want to add a small changeset so we can publish the doc changes with the npm package.

KrisQ:
Good call thank you @mdarrik, the link to the v1 documentation has been added!

PR #272, opened 2022-10-13
fix: changeset config/format

KrisQ:

Fixes [[ENG-7315](https://nacelle.atlassian.net/browse/ENG-7315)] 

- Configure nacelle-js changesets/action to use conventional commits and update lockfiles


package.json
```diff
@@ -15,6 +15,7 @@
     "test:ci": "jest --clearCache && lerna run test:ci",
     "changeset": "changeset",
     "postinstall": "lerna bootstrap",
+    "version": "changeset version && npm i --package-lock-only",
```
mdarrik, line 18: I think this will only update the lockfile for the root package-lock.json. We might need either a script or a lerna command to handle versioning each package's lock file (similar to how lerna lets you run tests for every child package that has a test command iirc). [Lerna 5 has these docs on it](https://lerna.js.org/docs/features/run-tasks).

NWRichmond (approved):
Thanks! Looks like the `lerna run --parallel lockfile:update` will take care of all of the `packages/**/package-lock.json` updates :)

PR #269, opened 2022-10-10
chore: merge main into shopify-cart branch

mdarrik:
## WHY are these changes introduced?

- This syncs changes on `main` with this feature branch. This allows us to have a `@nacelle/shopify-cart` version of the Gatsby reference store in this branch.

## WHAT is this pull request doing?

1. Merges main into the `shopify-cart-1.0.0` branch and handles the merge conflicts. Namely by deleting conflicting `package-lock.json` files and running `npm i --package-lock-only` to get the updated package-lock files.

## How to test

1. Checkout `mdarrik/merge-main-to-shopify-cart`
2. Run `npm run clean && CI=true npm run bootstrap` from the repo root - this makes sure there are no package-lock issues.

PR #274, opened 2022-10-18
chore: update lockfiles

NWRichmond:
The goal of this PR is to fix some lockfile issues introduced in #269, which appears to contain `npm link` symlinks in lockfiles.

Steps taken:
1. `npm run clean && rm -rf node_modules && npm i` from the monorepo root.
2. Upgrade `next`, `react`, and `react-dom` in `starters/next` and `reference-stores/next`. The impetus for this was Next Starter build failures [related to SWC](https://nextjs.org/docs/messages/failed-loading-swc)...something obscure was messed up, and it felt like a good opportunity to promote using the latest versions of these packages.


KrisQ (approved):
Thanks Nick!

PR #284, opened 2022-10-27
chore: add SonarCloud to shopify-checkout

KrisQ:

Fixes [[ENG-7228](https://nacelle.atlassian.net/browse/ENG-7228)] 

- configure SonarCloud for `shopify-checkout`

### Note:

Screen Shot 2022-10-27 at 2 56 43 PM


.gitignore
```diff
@@ -4,6 +4,7 @@ node_modules
 # build outputs
 dist
 build
+!.github/**/
```
NWRichmond, line 7: Do we need this?

NWRichmond (approved):
Looks good, @KrisQ! I was going to ask if we could move the `artiomtr/jest-coverage-report-action` job to `.github/workflows/pr.yml`, but we probably don't need it any longer now that coverage on PRs will be provided by `sonarsource/sonarcloud-github-action`.

Thanks for this PR :)

.gitignore
KrisQ: We don't anymore, thank you for catching that, and thank you for the review!

PR #278, opened 2022-10-20
feat: add `@nacelle/shopify-cart` package

KrisQ:
## Why are these changes introduced?

Addresses [ENG-6843](https://nacelle.atlassian.net/browse/ENG-6843)

> `@nacelle/shopify-cart` version `1.x`

## What is this pull request doing?

- This PR adds a new package, `@nacelle/shopify-cart@1.x`. Previously, the `@nacelle/shopify-cart@0.x` lived in a separate private repo.
- Updates the starters & reference stores to use `@nacelle/shopify-cart@1.x` instead of `@nacelle/shopify-checkout`.

### With `@nacelle/shopify-checkout`

Compared to `@nacelle/shopify-checkout`, `@nacelle/shopify-cart` offers the ability to leverage features that are specific to Shopify Carts and are not available in Shopify Checkouts. Examples include support for Shopify Selling Plans and Shopify Scripts discounts.

### With `@nacelle/shopify-cart@0.x`

Version `1.x` brings a variety of improvements, including the ability to customize the GraphQL requests made to the Shopify Storefront API. For a complete description of the changes, please see `packages/shopify-cart/CHANGELOG.md`.

## Prerequisites

- [x] https://github.com/getnacelle/nacelle-js/pull/299 must be merged before this PR.

PR #285, opened 2022-10-31
chore: add SonarCloud scan + coverage analysis

KrisQ:

Fixes [[ENG-7229](https://nacelle.atlassian.net/browse/ENG-7229)] 

- configure SonarCloud for shopify-cart


.github/workflows/release.yml
```diff
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - main
-      - ENG-6843/shopify-cart-1.0.0
```
NWRichmond, line 8: I don't think we can remove this until #278.

NWRichmond (changes requested):
Looks good! The only issue I see is that we can't yet remove `ENG-6843/shopify-cart-1.0.0` as a release branch, as doing so would prevent us from releasing `@beta` versions of `@nacelle/shopify-cart` before #278 is merged.

.github/workflows/release.yml
KrisQ: My bad, should be fixed now! Thank you @NWRichmond

NWRichmond (approved):
Looks good! Thanks for that update, @KrisQ.

PR #292, opened 2022-11-23
fix: allow plaintext Shopify global IDs

NWRichmond:

Fixes [ENG-7943](https://nacelle.atlassian.net/browse/ENG-7943)

> Shopify checkout ID verification doesn't account for plaintext IDs

Shopify [made a change](https://shopify.dev/changelog/improvements-and-changes-coming-to-the-storefront-api) to the Shopify Storefront API that creates a bug in `@nacelle/shopify-checkout`. Previously, Shopify returned base64-encoded global IDs for all nodes (including Checkout nodes).

[Shopify’s description of this change includes](https://shopify.dev/changelog/changes-to-storefront-api-object-ids):

> …when requesting the unstable, 2022-04 release candidate, or future versions of the Storefront API, the object IDs returned will no longer be Base64 encoded.

This change breaks [@nacelle/shopify-checkout's `isVerifiedCheckoutId` utility function](https://github.com/getnacelle/nacelle-js/blob/main/packages/shopify-checkout/src/utils/isVerifiedCheckoutId.ts), which currently expects that all global IDs are Base64-encoded.

This bug flew under the radar because the `starters/*` and `reference-stores/*` projects don't supply an `id` to `checkoutClient.process`. Arguably, they should, when a checkout `id` has been persisted.

- Updates the `isVerifiedCheckoutId` utility function to accept either plaintext or base64-encoded checkout IDs.
  - Also updates associated tests & mocks.
- Removes `storefrontApiVersion` language from the README to avoid confusion. This parameter was removed in a previous version of `@nacelle/shopify-checkout`.

## How to Test

First, the basics:

1. Check out the `ENG-7943-shopify-checkout-allow-plaintext-ids` branch.
2. Navigate to `packages/shopify-checkout`
3. `npm run build` - the package should build without issues.
4. `npm run test:ci` - all tests should pass without a decrease in coverage.

Next, I'd recommend taking this change for a spin in one of the Reference Stores. I used the Nuxt Reference Store, and found that:

- [x] checkout creation & updates worked as expected.
- [x] when passing an `id` to `processCheckout`, there were no errors, and `checkoutId` was included in the request.

https://user-images.githubusercontent.com/5732000/203462840-859a4bd9-d0f9-4dc8-acc2-ba8f9a379e48.mov

Here are the steps I followed, in addition to "the basics" described above:

1. Navigate to `reference-stores/nuxt`
2. `npm i @nacelle/shopify-checkout@latest`
3. Navigate back to the monorepo root
4. `npm run bootstrap`
5. Navigate back to  `reference-stores/nuxt`
6. In `reference-stores/nuxt/store/checkout.js`'s `processCheckout` action, add the following after line 45, modify the `checkoutClient.process` call to include the persisted checkout ID:
```js
const checkoutId = await get('checkoutId');
const checkoutData = await checkoutClient.process({
  id: checkoutId,
  cartItems
});
```
7. `npm run dev` and visit [`localhost:3000`](http://localhost:3000)
8. Add an item to the cart
9. Initiate a checkout, but do not complete the checkout
10. Hit the back button to return to [`localhost:3000`](http://localhost:3000)
11. In devtools, check the **Application** tab and inspect the IndexDB store. The `checkoutId` should be a plaintext ID.
12. Add more items to the cart
13. Initiate a checkout. The `checkoutId` should be included in the request payload, and there should be no errors.

## Next steps

While troubleshooting this issue, it became clear that our monorepo versioning setup isn't bumping versions as expected. For example, in https://github.com/getnacelle/nacelle-js/pull/291, the version of `@nacelle/shopify-checkout` was bumped, but it was not bumped in the `"dependencies"` of the various starters & reference stores' `package.json`s.

### DIFF
--- a/packages/shopify-checkout/__tests__/mocks/index.ts
+++ b/packages/shopify-checkout/__tests__/mocks/index.ts
@@ -24,16 +24,15 @@ const checkoutUuids = {
 export const checkoutIds = {
   encoded: {
     beginsWithLetter: Buffer.from(
-      'gid://shopify/Checkout/' + checkoutUuids.beginsWithLetter
+      `gid://shopify/Checkout/${checkoutUuids.beginsWithLetter}`
     ).toString('base64'),
     beginsWithNumber: Buffer.from(
-      'gid://shopify/Checkout/' + checkoutUuids.beginsWithNumber
+      `gid://shopify/Checkout/${checkoutUuids.beginsWithNumber}`
     ).toString('base64')
   },
   plaintext: {
-    beginsWithLetter:
-      'gid://shopify/Checkout/' + checkoutUuids.beginsWithLetter,
-    beginsWithNumber: 'gid://shopify/Checkout/' + checkoutUuids.beginsWithNumber
+    beginsWithLetter: `gid://shopify/Checkout/${checkoutUuids.beginsWithLetter}`,
+    beginsWithNumber: `gid://shopify/Checkout/${checkoutUuids.beginsWithNumber}`
   }
 };
 
--- a/packages/shopify-checkout/__tests__/mocks/index.ts
+++ b/packages/shopify-checkout/__tests__/mocks/index.ts
@@ -22,12 +22,19 @@ const checkoutUuids = {
   beginsWithNumber: '998877?key=123123'
 };
 export const checkoutIds = {
-  beginsWithLetter: Buffer.from(
-    'gid://shopify/Checkout/' + checkoutUuids.beginsWithLetter
-  ).toString('base64'),
-  beginsWithNumber: Buffer.from(
-    'gid://shopify/Checkout/' + checkoutUuids.beginsWithNumber
-  ).toString('base64')
+  encoded: {
+    beginsWithLetter: Buffer.from(
+      'gid://shopify/Checkout/' + checkoutUuids.beginsWithLetter
+    ).toString('base64'),
+    beginsWithNumber: Buffer.from(
+      'gid://shopify/Checkout/' + checkoutUuids.beginsWithNumber
+    ).toString('base64')
+  },
+  plaintext: {
+    beginsWithLetter:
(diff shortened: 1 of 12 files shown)


KrisQ (approved):
Looks good + works great!

Screen Shot 2022-11-23 at 10 42 06 AM
Screen Shot 2022-11-23 at 11 04 20 AM
Screen Shot 2022-11-23 at 11 16 10 AM

packages/shopify-checkout/__tests__/mocks/index.ts
```diff
@@ -22,12 +22,19 @@ const checkoutUuids = {
...
 export const checkoutIds = {
-  beginsWithLetter: Buffer.from(
-    'gid://shopify/Checkout/' + checkoutUuids.beginsWithLetter
-  ).toString('base64'),
-  beginsWithNumber: Buffer.from(
-    'gid://shopify/Checkout/' + checkoutUuids.beginsWithNumber
-  ).toString('base64')
+  encoded: {
+    beginsWithLetter: Buffer.from(
+      'gid://shopify/Checkout/' + checkoutUuids.beginsWithLetter
```
mdarrik, line 27: Not something we _need_ to fix in this PR, but we might want to consider using template literals here instead of concatenation. I usually find them cleaner/easier to read.

NWRichmond: No problem at all, added in https://github.com/getnacelle/nacelle-js/pull/292/commits/5dbe9a252c96cced6660dfa4f3bc1ed0a30bbe99 :)

mdarrik (approved):
Looks good! Tested with the next reference store + some code to preserve & use the checkout Id during process checkout, and with the old version it errored. With the new version it worked perfectly. Thanks so much!

NWRichmond:
Awesome! Thank you for the timely & thorough reviews @mdarrik & @KrisQ 🤝

PR #301, opened 2022-12-14
docs: cleanup shopify cart changelogs

mdarrik:
## Cleanup the reference stores + starter changelogs

This is a small PR to cleanup the reference store/starter PR's, which feature a bunch of `Update dependencies` methods in their changelogs for a single version bump. 

We lose the commit list of dependency updates, but we also have those in PRs too. 

Also happy to close this if we want to just merge them as is.


NWRichmond (approved):
Looks good!

PR #304, opened 2022-12-22
feat: generate TS types from Storefront GraphQL

NWRichmond:

Addresses [ENG-8184](https://nacelle.atlassian.net/browse/ENG-8184)

> Configure GraphQL codegen

This PR configures [GraphQL Code Generator](https://the-guild.dev/graphql/codegen) to generate the following TS types:
- All types available in the [Nacelle Storefront API](https://docs.nacelle.com/docs/storefront-graphql-api-stable)
- Types for the Nacelle Storefront API results of GraphQL fragments and queries in `packages/storefront-sdk/src/graphql`

### Discussion

There are some noteworthy aspects to the changes introduced by this PR:

#### Node.js x ESM x TS

In keeping with [GraphQL Codegen's documentation for TypeScript x ESM](https://the-guild.dev/graphql/codegen/docs/getting-started/esm-typescript-usage#typescript-compiler-options), the TypeScript `moduleResolution` and `module` config have been updated to `"Node16"`. This change, coupled with the fact that this package uses `"type": "module"` requires us to handle imports a bit differently than we're used to. Instead of:

```ts
import { thing } from '../someFile';
```

, we must instead import using a `.js` extension:

```ts
import { thing } from '../someFile.js';
```

You can learn more about why this is [in the TypeScript 4.7 Release Notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#ecmascript-module-support-in-node-js).

The two consequences of this change are:
1. We need to get used to importing with the `.js` extension.
2. The ESLint config has been updated with the requisite `extensionAlias` settings.

For some lively reading on why we can't use a `.ts` extension instead of a `.js` extension for ESM imports, see https://github.com/microsoft/TypeScript/issues/49083.

#### The `graphql` dependency

`graphql` has been installed as a dependency. This is probably fine, but it's worth contemplating whether this could lead to multiple versions of `graphql` being installed in a user's project, which can cause some tricky issues. I'm not sure how likely that would be. But if we can identify specific cases where it would be a problem, one potential solution would be to reinstall `graphql` as a devDep, and also include it in `peerDependencies`.

### How to Test

1. Check out the `ENG-8184-configure-graphql-codegen` branch.
4. `npm run bootstrap` from the monorepo root, or `npm i` from `packages/storefront-sdk`.
5. Navigate to `packages/storefront-sdk`.
6. `npm run build` - the package should build without errors.
7. `npm run codegen` - the codegen steps should finish without errors, and there should be no uncommitted changes to `src/types/storefront.ts`.


packages/storefront-sdk/codegen.yml
```diff
@@ -0,0 +1,10 @@
+overwrite: true
+schema: 'https://storefront.api.nacelle.com/graphql/v1/spaces/${NACELLE_SPACE_ID}'
+documents: 'src/graphql/**/*.ts'
+generates:
+  src/types/storefront.ts:
+    plugins:
+      - typescript
+      - typescript-operations
+    config:
+      enumsAsTypes: true
```
NWRichmond, line 10: If you're wondering why we're using `codegen.yml` instead of `codegen.ts`, despite what [the GraphQL Code Generator docs](https://the-guild.dev/graphql/codegen/docs/getting-started/esm-typescript-usage) say, there is a bug that prevents us from using `codegen.ts` in an ESM project: https://github.com/dotansimha/graphql-code-generator/issues/8488.

KrisQ (approved):
Looks good!

mdarrik (approved):
LGTM! Even after deleting `src/types/storefront.ts` and re-running it, there are no changes to be committed.

PR #306, opened 2022-12-30
feat: add basic storefront client

NWRichmond:

Addresses [ENG-8318](https://nacelle.atlassian.net/browse/ENG-8318)

> Scaffold the createStorefrontClient function and StorefrontClient class

This PR adds a basic implementation of the `Storefront` client creation function and a `StorefrontClient` class that encapsulates both the public API and private implementation details.

## How to Test

1. Check out the `ENG-8318-storefront-client-scaffolding` branch.
2. Navigate to `packages/storefront-sdk`.
3. `npm run build` - it should create `dist/nacelle-storefront-sdk.js` and `dist/nacelle-storefront-sdk.umd.cjs` without errors.
4. `npm run coverage` - all tests should pass with 100% coverage.

Storefront SDK all test passing with 100 percent coverage ENG-8318-storefront-client-scaffolding branch

## Testing checklist

### DIFF
--- a/packages/storefront-sdk/src/client/index.ts
+++ b/packages/storefront-sdk/src/client/index.ts
@@ -3,10 +3,10 @@ import type { Client as UrqlClient } from '@urql/core';
 import type { StorefrontClientParams } from '../index.js';
 
 export class StorefrontClient {
-	private readonly graphqlClient: UrqlClient;
+	readonly #graphqlClient: UrqlClient;
 
 	constructor(params: StorefrontClientParams) {
-		this.graphqlClient = createClient({
+		this.#graphqlClient = createClient({
 			url: params.storefrontEndpoint,
 			fetch: params.fetchClient ?? globalThis.fetch
 		});
@@ -14,6 +14,6 @@ export class StorefrontClient {
 
 	// NEW METHODS GO HERE :)
 	placeholder() {
-		return this.graphqlClient.query('', {}).toPromise();
+		return this.#graphqlClient.query('', {}).toPromise();
 	}
 }
--- a/packages/storefront-sdk/src/client/index.ts
+++ b/packages/storefront-sdk/src/client/index.ts
@@ -1,0 +1,19 @@
+import { createClient } from '@urql/core';
+import type { Client as UrqlClient } from '@urql/core';
+import type { StorefrontClientParams } from '../index.js';
+
+export class StorefrontClient {
+	private readonly graphqlClient: UrqlClient;
+
+	constructor(params: StorefrontClientParams) {
+		this.graphqlClient = createClient({
+			url: params.storefrontEndpoint,
+			fetch: params.fetchClient ?? globalThis.fetch
+		});
+	}
+
+	// NEW METHODS GO HERE :)
+	placeholder() {
+		return this.graphqlClient.query('', {}).toPromise();
+	}
+}
--- a/packages/storefront-sdk/src/index.ts
+++ b/packages/storefront-sdk/src/index.ts
@@ -1,7 +1,18 @@
-function placeholder(): object {
-	console.log('placeholder');
+import { StorefrontClient } from './client/index.js';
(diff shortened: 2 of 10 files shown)


NWRichmond:
(I don't think we need a changeset to capture these changes)

packages/storefront-sdk/src/index.ts
```diff
@@ -1,7 +1,18 @@
...
 }
 
-export default placeholder;
+export function Storefront(params: StorefrontClientParams) {
+	if (!params?.storefrontEndpoint) {
+		throw new Error(errorMessages.missingEndpoint);
+	}
+
+	return new StorefrontClient(params);
+}
```
mdarrik, line 18: Do we want to also export this function as the default to match [the v1 sdk](https://github.com/getnacelle/nacelle-storefront-sdk/blob/80626c98938ec3107b6dc9ec31d06ae91b76489a/src/index.ts#L7)?

packages/storefront-sdk/src/client/index.ts
```diff
@@ -0,0 +1,19 @@
+import { createClient } from '@urql/core';
+import type { Client as UrqlClient } from '@urql/core';
+import type { StorefrontClientParams } from '../index.js';
+
+export class StorefrontClient {
+	private readonly graphqlClient: UrqlClient;
```
mdarrik, line 6: Wonder if we should consider the native JS private feature? 

When using the node repl or a console log, this is what the output looks like for each version: 

TS Private keyword - the output is also the same as console.log
```sh
> testClient = client({storefrontEndpoint: 'http://localhost:8888'})
Ln {
  graphqlClient: e {
    suspense: false,
    'operations$': [Function (anonymous)],
    reexecuteOperation: [Function: reexecuteOperation],
    createRequestOperation: [Function: createRequestOperation],
    executeRequestOperation: [Function: executeRequestOperation],
    executeQuery: [Function: executeQuery],
    executeSubscription: [Function: executeSubscription],
    executeMutation: [Function: executeMutation],
    query: [Function: query],
    readQuery: [Function: readQuery],
    subscription: [Function: subscription],
    mutation: [Function: mutation],
    subscribeToDebugTarget: [Function (anonymous)]
  }
}
```
js native private field with `#`
```sh
testClient = client({storefrontEndpoint: 'http://localhost:8888'})
Fn {}
```

Also worth noting that the node repl gives me autocomplete on the TS private version, but doesn't on the native one.

packages/storefront-sdk/src/index.ts
NWRichmond: That's certainly a possibility, although I wouldn't mind making the removal of the default export a breaking change. When we use the default export, we get this warning from Vite, as we've seen with other Vite projects with mixed exports:

Entry module  src:index ts  is using named and default exports together  Consumers of your bundle will have to use `NacelleStorefrontSdk default` to access the default export, which may not be what you want

packages/storefront-sdk/src/client/index.ts
NWRichmond: The way that I've been thinking about private fields is not so much "make it so that this field can't be overwritten" but more "make it so that this field doesn't appear in Intellisense." But if you feel there's value in making it [hard private](https://www.typescriptlang.org/docs/handbook/2/classes.html#caveats), I'm happy to make that change.

NWRichmond: Implemented in https://github.com/getnacelle/nacelle-js/pull/306/commits/7770cd4370b5d6874fb99da5932bc1b0bf2a81ed

packages/storefront-sdk/src/index.ts
NWRichmond: Per conversation on Zoom, we'll leave this as a named export for now, but will explore impact of this breaking change and leave the default export as a possibility going forward.

KrisQ (approved):
Looks good!

PR #307, opened 2023-01-04
feat: add v2 query method using urql

mdarrik:
- feat: configure typed-document-node generation
- test: configure vitest typechecking
- test: configure request mocks
- feat: adds v2 query method
- docs: adds changeset for query method changes

## WHY are these changes introduced?

Addresses [ENG-8327](https://nacelle.atlassian.net/browse/ENG-8327) - Adds the `.query` method to the v2 of the storefront-sdk. 

## WHAT is this pull request doing?

1. Adds support for TypedDocumentNodes so we can get typed query responses & users can get typed responses. Updates `codegen` to generate the typed queries from any queries we include in the sdk's `src/graphql/` folder.
2. Initializes the `query` method of the v2 sdk. Note that unlike v1, we return `{data, error}` from the method so that users can handle errors without needing to use `catch`. This is a breaking change between v1 and v2.
3. Adds infrastructure for mocking Requests including a mocked Navigation response.
4. Sets up vitest's test type checking & runs it as part of the `coverage` command. This stops us from being able to use dynamic tests like `each`, but lets us validate types in our tests.

## How to Test

1. Checkout this branch `mdarrik/ENG-8327-add-sdk-v2-query-method`
2. Make sure you're using Node 18 (new tests rely on )
3. Install dependencies by running `npm run bootstrap` at the root of the repo (there were dependency changes with this PR)
4. Run the tests with `npm run coverage` - the typechecking test and the coverage test should both pass
5. Build the package by running `npm run build`, should succeed
6. Test the new `coverage`
7. Test the `.query` method by running either the node repl `node` or using `npm link` and then `npm link @nacelle/storefront-sdk` and using the method in another npm project. Can use the [TypedDocumentNode instructions](https://github.com/dotansimha/graphql-typed-document-node#how-to-use) to generate typed queries/responses.

[ENG-8327]: https://nacelle.atlassian.net/browse/ENG-8327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-8327]: https://nacelle.atlassian.net/browse/ENG-8327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #310, opened 2023-01-04
feat: implement the `after` method

NWRichmond:

Addresses [ENG-8320](https://nacelle.atlassian.net/browse/ENG-8320) 

> Implement the after method

This PR implements`@nacelle/storefront-sdk`'s [`.after`](https://docs.nacelle.com/docs/javascript-storefront-sdk#transforming-the-responses-of-sdk-methods) method.

This particular implementation introduces some non-breaking changes.

### Callbacks are registered by ID & deletable

In `@nacelle/storefront-sdk@1.x`, callbacks registered with the `.after` method were not deletable. This PR changes how callbacks are stored - rather than an array, callbacks are stored in an object that is namespaced by method. If the user provides an optional `callbackId`, we use that as the key. Otherwise, the key takes the form:

```ts
`${methodName}::${n}`
```

where `n` is the current number of callbacks registered to a particular method.

Storing callbacks this way allows them to be deleted. Users can delete a callback by calling the `.after` method with the `callbackId` of interest + `callback: null`:

```ts
// Register a callback
client.data.after(
  'spaceProperties',
  (data: SpaceProperties) => data,
  'myCallback'
);

// Delete the callback
client.data.after('spaceProperties', null, 'myCallback');
```

### The type for the `callback` supplied to `.after` reacts to the `method` name

Unlike `@nacelle/storefront-sdk@1.x`, the `.after` method implementation in this PR provides strong typing for the `callback` supplied to `.after`. The type is determined based on the `method` name provided to `.after`. For example:

https://user-images.githubusercontent.com/5732000/210666640-ba365c6b-5412-4827-bb7f-e15c75b4d6c5.mov

### Other improvements

The `method` supplied to `.after` is a string union of methods instead of a `string`. This gives a nice autocomplete experience in VSCode when using TypeScript.

### Next steps and open questions

- After #307 is merged, we can add unit tests for `applyAfter`.
- As noted in the comments, there are a handful of places that will likely need revision after we have a clearer concept of the plugin system and whether data-fetching methods other than `.query` will be provided via plugins.

## How to Test

1. Check out the `ENG-8320-after-method` branch.
2. Navigate to `packages/storefront-sdk`.
3. `npm run build` - the package should build without errors.
4. `npm run coverage` - all tests should pass, and the only uncovered lines should be for the `applyAfter` method. I think it makes sense for this method to be tested with `.query`, as follow-up to this PR and #307.

ENG-8320-after-method branch all tests passing some lines not covered in src:client:index

## Checklist

- [x] This Pull Request aligns with `nacelle-js`' [Code of Conduct](https://github.com/getnacelle/nacelle-js/blob/main/CODE_OF_CONDUCT.md#code-of-conduct).
- [x] Screenshots, videos, etc. in the PR description are free of brand names and sensitive details.

[ENG-8320]: https://nacelle.atlassian.net/browse/ENG-8320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-8320]: https://nacelle.atlassian.net/browse/ENG-8320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-8320]: https://nacelle.atlassian.net/browse/ENG-8320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/storefront-sdk/src/client/index.ts
+++ b/packages/storefront-sdk/src/client/index.ts
@@ -15,11 +15,18 @@ import type {
 	DataFetchingMethodName,
 	MethodData
 } from '../types/after.js';
-import type { Content } from '../types/storefront.js';
 
 export interface StorefrontResponse<QueryDocumentType> {
 	error?: CombinedError;
 	data?: QueryDocumentType;
+}
+
+export interface QueryParams<QData, QVariables extends AnyVariables> {
+	/** GraphQL query */
+	query: TypedDocumentNode<QData, QVariables> | DocumentNode | string;
+
+	/** GraphQL query variables */
+	variables?: QVariables | string;
 }
 
 export class StorefrontClient {
@@ -121,14 +128,10 @@ export class StorefrontClient {
 	query<QData = any, QVariables extends AnyVariables = any>({
 		query,
 		variables
-	}: {
-		query: TypedDocumentNode<QData, QVariables> | DocumentNode | string;
-		variables?: QVariables | string;
-	}): Promise<StorefrontResponse<QData>> {
+	}: QueryParams<QData, QVariables>): Promise<StorefrontResponse<QData>> {
 		return this.#graphqlClient
 			.query(query, variables as QVariables)
 			.toPromise()
-			.then(({ data, error }) => ({ data, error }));
+			.then(({ data, error }) => this.applyAfter('query', { data, error }));
 	}
-	/* c8 ignore stop */
 }
--- a/packages/storefront-sdk/src/client/index.ts
+++ b/packages/storefront-sdk/src/client/index.ts
@@ -1,4 +1,5 @@
 import { createClient } from '@urql/core';
+import { errorMessages, dataFetchingMethods } from '../utils/index.js';
 import type {
 	Client as UrqlClient,
 	TypedDocumentNode,
@@ -7,6 +8,14 @@ import type {
 } from '@urql/core';
 import type { DocumentNode } from 'graphql';
 import type { StorefrontClientParams } from '../index.js';
+import type {
+	AfterCallback,
+	AfterCallbackWithId,
+	AfterSubscriptions,
+	DataFetchingMethodName,
+	MethodData
+} from '../types/after.js';
+import type { Content } from '../types/storefront.js';
 
 export interface StorefrontResponse<QueryDocumentType> {
 	error?: CombinedError;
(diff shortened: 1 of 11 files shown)


packages/storefront-sdk/src/client/index.ts
```diff
@@ -1,19 +1,125 @@
...
+
+		if (typeof methodSubscriptions === 'undefined') {
+			methodSubscriptions = this.#afterSubscriptions[method] = {};
+		}
+
+		const id =
+			callbackId ?? `${method}::${Object.values(methodSubscriptions).length}`;
+
+		if (typeof callback === 'function') {
+			(methodSubscriptions as unknown as AfterCallbackWithId<MethodName>)[id] =
```
NWRichmond, line 73: This type assertion bothers me a little. But without it, TypeScript seems unable to understand what the type should be.

packages/storefront-sdk/src/types/after.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export type MethodData = {
+	products: Product[] | ProductEdge[];
+	productCollections: ProductCollection[] | ProductCollectionEdge[];
+	productCollectionEntries: Product[] | ProductEdge[];
+	content: Content[] | ContentEdge[];
+	navigation: NavigationGroup[];
+	spaceProperties: SpaceProperties;
+	query: Record<string, unknown> | Array<Record<string, unknown>>;
+};
```
NWRichmond, line 25: We probably will need to rethink this once we have a clearer picture of how plugins will work, and whether these methods (other than `.query`) will be made available by a plugin.

packages/storefront-sdk/src/utils/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const dataFetchingMethods = [
+	'content',
+	'navigation',
+	'products',
+	'productCollections',
+	'productCollectionEntries',
+	'query',
+	'spaceProperties'
+] as const;
```
NWRichmond, line 9: We probably will need to rethink this once we have a clearer picture of how plugins will work, and whether these methods (other than `.query`) will be made available by a plugin.

KrisQ (approved):
Looks good!

Screen Shot 2023-01-04 at 4 59 39 PM

packages/storefront-sdk/src/client/index.ts
mdarrik: It seems like we should be able to do away with this assertion at some stage. This is the error TS gives, which makes me think that there's something little we can fix since it's just acting like the two types don't match despite having virtually identical signatures: 

```sh
 Type 'AfterCallback<MethodData[MethodName]>' is not assignable to type 'AfterCallback<Content[] | ContentEdge[] | Product[] | ProductCollectionEdge[] | ProductEdge[] | NavigationGroup[] | ProductCollection[] | SpaceProperties | StorefrontResponse<...>>'.
  Types of parameters 'responseObj' and 'responseObj' are incompatible.
    Type 'Content[] | ContentEdge[] | Product[] | ProductCollectionEdge[] | ProductEdge[] | NavigationGroup[] | ProductCollection[] | SpaceProperties | StorefrontResponse<...>' is not assignable to type 'MethodData[MethodName]'.
```

NWRichmond: At least part of the problem seems to be that TS doesn't know that `this.#afterSubscriptions[method]` cannot be `undefined` on this line, despite the assignment above:

```ts
if (typeof methodSubscriptions === 'undefined') {
	methodSubscriptions = this.#afterSubscriptions[method] = {};
}
```

If you remove the assertion but keep parentheses, we get:

```
Object is possibly 'undefined'. ts(2532)
```

at the `(methodSubscriptions)[id]` assignment.

object is possibly undefined

It might have something to do with https://github.com/microsoft/TypeScript/issues/27706.

mdarrik: It seems like there's something subtly off about the assignment type for sure. With some non-null assertions and string type casts to make sure it recognize the id type correctly, I still get this: 

```sh
Type 'AfterCallback<MethodData[MethodName]>' is not assignable to type 'NonNullable<Partial<Record<"query" | "navigation" | "content" | "products" | "productCollections" | "productCollectionEntries" | "spaceProperties", AfterCallbackWithId<"query" | "navigation" | "content" | "products" | "productCollections" | "productCollectionEntries" | "spaceProperties">>>[MethodName]>[MethodName]'
```

mdarrik (approved):
Seems good! Tests all pass and I'm able to use the after callback locally. Thanks for this!

PR #344, opened 2023-02-27
chore: remove request policy

KrisQ:

Addresses [ENG-8873](https://nacelle.atlassian.net/browse/ENG-8873) 

`requestPolicy: 'network-only'` has no effect without the cache exchange and would prevent the user from setting their request policy.

Removes the request policy from the `StorefrontClient` configuration.


NWRichmond (approved):
Great! Thanks for this PR, @KrisQ 🤝

PR #354, opened 2023-03-09
refactor: make `after` method plugin-agnostic

NWRichmond:

Addresses [ENG-8895](https://nacelle.atlassian.net/browse/ENG-8895)

> Refactor after/applyAfter to not conflate commerce queries with core SDK

This PR refactors the `after` method to decouple it from the Commerce Queries plugin. Previously, the `after` method relied on knowledge of the return types of Commerce Queries methods. This was not sustainable (the core SDK cannot be aware of the function signatures of every method of every future Storefront SDK plugin).

The implementation of the `after` method in this PR is closer to Storefront SDK `1.x` than prior `2.0.0-beta.x` versions. The disadvantage to this approach is that we lose type safety - it is much easier to write an `after` callback that creates runtime errors due to a mismatch between the callback and the method that it's applied to. Unfortunately, it doesn't seem like there is a way around this, even with conditional types.

## Checklist

- [ ] This Pull Request aligns with `nacelle-js`' [Code of Conduct](https://github.com/getnacelle/nacelle-js/blob/main/CODE_OF_CONDUCT.md#code-of-conduct).
- [ ] You can follow your own "How to Test" instructions and get the expected result.
- [ ] Screenshots, videos, etc. in the PR description are free of brand names and sensitive details.
- [ ] Created a [changeset](https://github.com/changesets/changesets) (if the change should appear in a changelog).
- [ ] Updated the `README.md` with documentation changes (if applicable).


KrisQ (approved):
LGTM!

Screen Shot 2023-03-09 at 9 54 35 AM
Screen Shot 2023-03-09 at 9 54 05 AM

PR #383, opened 2023-04-27
build: skip lockfile updates

NWRichmond:
This PR temporarily disables lockfile updates in the **Release** script to overcome issues with https://github.com/getnacelle/nacelle-js/actions/runs/4822366925/jobs/8589502929


KrisQ (dismissed):

norrismiv (changes requested):
These lines will need to be commented out as well. 60-64

norrismiv (approved):

PR #409, opened 2023-11-02
fix: using typename in section

KrisQ:

Addresses [ENG-8700](https://nacelle.atlassian.net/browse/ENG-8700) 

- Update TypedFields names to reflect the new name in the schema
- Update the sections to use `__typename` and a `switch` statement to render components
- Remove unecessary `__typename` check in Article and Links components 

## How to Test

`npm run dev` in `examples/typedfields`

## Checklist

- [x] This Pull Request aligns with `nacelle-js`' [Code of Conduct](https://github.com/getnacelle/nacelle-js/blob/main/CODE_OF_CONDUCT.md#code-of-conduct).
- [x] Screenshots, videos, etc. in the PR description are free of brand names and sensitive details.
- [ ] Created a [changeset](https://github.com/changesets/changesets) (if the change should appear in a changelog).
- [ ] Updated the `README.md` with documentation changes (if applicable).

[ENG-8700]: https://nacelle.atlassian.net/browse/ENG-8700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NWRichmond:
Nice! Everything's working as expected.

But...I think perhaps I wasn't clear about something when we were discussing `<Section />` yesterday. While the `<Section />` component with the `switch` definitely works in this case, I don't think it's a pattern that customers should take inspiration from, because in a "real" project it will import many unused components into a page that `<Section />` is used in.

Instead, I think we can do without with `<Section />` component altogether, and add the `switch` handling to the page template. That way, any page template can define which components it will use. It shouldn't make a bit of difference in this particular project, but I think it might be helpful to people using this project for inspiration.

Please let me know if you disagree, @KrisQ!

KrisQ:
Thank you for clarifying @NWRichmond, it was a pretty simple change that should now be in!

NWRichmond (approved):
Beautiful! Thank you for this, @KrisQ 🤝

